PR #15, opened 2021-10-06
Adding new field to SDK constructor to accept env var

Dmi3D:
# Why
Cage playground in staging cannot use the React SDK because the production API URL is hard coded in the Javascript SDK config.

# How
Adding a path to the constructor to accept env var as the keys url path, and passing that through to the config.

### DIFF
--- a/lib/v1/config.js
+++ b/lib/v1/config.js
@@ -1,7 +1,7 @@
 const KEYS_URL = 'https://keys.evervault.com';
 const METRICS_URL = ' https://metrics.evervault.com';
 
-module.exports = (teamId) => ({
+module.exports = (teamId, keysUrlOverride) => ({
   teamId,
   encryption: {
     cipherAlgorithm: 'aes-256-gcm',
@@ -14,7 +14,7 @@ module.exports = (teamId) => ({
     },
   },
   http: {
-    keysUrl: KEYS_URL,
+    keysUrl: keysUrlOverride || KEYS_URL,
     metricsUrl: METRICS_URL,
   }
 })
--- a/lib/v1/config.js
+++ b/lib/v1/config.js
@@ -1,11 +1,11 @@
 const KEYS_URL = 'https://keys.evervault.com';
 const METRICS_URL = 'https://metrics.evervault.com';
 
-module.exports = DEFAULT_CONFIG_URLS = {
+const DEFAULT_CONFIG_URLS = {
   keysUrl: KEYS_URL
 };
 
-module.exports = (teamId, urls = DEFAULT_CONFIG_URLS) => ({
+module.exports = (teamId, urlOverrides = DEFAULT_CONFIG_URLS) => ({
   teamId,
   encryption: {
     cipherAlgorithm: 'aes-256-gcm',
@@ -18,7 +18,7 @@ module.exports = (teamId, urls = DEFAULT_CONFIG_URLS) => ({
     },
   },
   http: {
-    keysUrl: urls.keysUrl || DEFAULT_CONFIG_URLS.keysUrl,
+    keysUrl: urlOverrides.keysUrl || DEFAULT_CONFIG_URLS.keysUrl,
     metricsUrl: METRICS_URL
   }
 })
--- a/lib/v1/config.js
+++ b/lib/v1/config.js
@@ -2,7 +2,7 @@ const KEYS_URL = 'https://keys.evervault.com';
 const METRICS_URL = 'https://metrics.evervault.com';
 
 const DEFAULT_CONFIG = {
-  urls = {
+  urls: {
     keysUrl: KEYS_URL
   }
 };
--- a/lib/v1/config.js
+++ b/lib/v1/config.js
@@ -21,6 +21,6 @@ module.exports = (teamId, config = DEFAULT_CONFIG) => ({
   },
   http: {
     keysUrl: config.urls.keysUrl,
-    metricsUrl = METRICS_URL
+    metricsUrl: METRICS_URL
   }
 })
--- a/lib/v1/config.js
+++ b/lib/v1/config.js
@@ -1,7 +1,12 @@
 const KEYS_URL = 'https://keys.evervault.com';
 const METRICS_URL = ' https://metrics.evervault.com';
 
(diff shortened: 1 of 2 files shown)


lib/v1/index.js
```diff
@@ -4,11 +4,14 @@ const { Crypto, Http, Forms } = require('./core');
...
 class EvervaultClient {
-  constructor(teamId) {
+  constructor(teamId, apiUrlOverride) {
     if (!Datatypes.isString(teamId)) {
       throw new errors.InitializationError('teamId must be a string');
     }
-    this.config = Config(teamId);
+    if (!Datatypes.isUndefined(apiUrlOverride) || !Datatypes.isString(apiUrlOverride)) {
+      throw new errors.InitializationError('appUrl must be a string');
+    }
```
lfarrel6, line 13: This is an optional argument, we don't want this to be set in general so shouldn't be erroring.

lfarrel6 (changes requested):

lib/v1/config.js
```diff
@@ -1,7 +1,12 @@
 const KEYS_URL = 'https://keys.evervault.com';
-const METRICS_URL = ' https://metrics.evervault.com';
+const METRICS_URL = 'https://metrics.evervault.com';
 
-module.exports = (teamId) => ({
+const DEFAULT_CONFIG_URLS = {
+	keysUrl: KEYS_URL,
+  metricsUrl: METRICS_URL
```
lfarrel6, line 6: We can omit this URL from here

lib/v1/index.js
```diff
@@ -1,14 +1,13 @@
 const { Datatypes, errors, extractDomain } = require('./utils');
 const Config = require('./config');
 const { Crypto, Http, Forms } = require('./core');
-const extractDomainFromUrl = require('./utils/domain');
 
 class EvervaultClient {
-  constructor(teamId) {
+  constructor(teamId, configUrls) {
```
lfarrel6, line 6: This should have a more generic name, to allow us to extend it. 
```suggestion
  constructor(teamId, config) {
```

lib/v1/index.js
```diff
@@ -1,14 +1,13 @@
...
     if (!Datatypes.isString(teamId)) {
       throw new errors.InitializationError('teamId must be a string');
     }
-    this.config = Config(teamId);
+    this.config = Config(teamId, configUrls);
```
lfarrel6, line 10: Only hand off the urls that we're setting to the config for now. (Will need some type checks to make sure `config` is defined and fallback to an empty object if not
```suggestion
    this.config = Config(teamId, config.urls);
```

lib/v1/config.js
```diff
@@ -13,8 +18,5 @@ module.exports = (teamId) => ({
       version: 1,
     },
   },
-  http: {
-    keysUrl: KEYS_URL,
-    metricsUrl: METRICS_URL,
-  }
+  http: configUrls
```
lfarrel6, line 21: configUrls can be anything here including undefined which would break the SDK. We need to add a check to ensure that it provides the expected urls otherwise we should fallback to the defaults

lib/v1/config.js
```diff
@@ -1,7 +1,12 @@
...
-const METRICS_URL = ' https://metrics.evervault.com';
+const METRICS_URL = 'https://metrics.evervault.com';
 
-module.exports = (teamId) => ({
+const DEFAULT_CONFIG_URLS = {
+	keysUrl: KEYS_URL,
+  metricsUrl: METRICS_URL
+};
+
+module.exports = (teamId, configUrls = DEFAULT_CONFIG_URLS) => ({
```
jakekgrog, line 9: Rather than returning the object straight away, you should add some checks that `configUrls` contains the expected fields/types and fall back to default values if not

jakekgrog (changes requested):

lib/v1/config.js
```diff
@@ -14,7 +20,7 @@ module.exports = (teamId) => ({
     },
   },
   http: {
-    keysUrl: KEYS_URL,
-    metricsUrl: METRICS_URL,
+    keysUrl: config.urls.keysUrl,
+    metricsUrl = METRICS_URL
```
eoinboylan, line 24: This isn't valid JS is it?

Dmi3D: 🤦‍♂️ thanks

lib/v1/index.js
```diff
@@ -1,14 +1,18 @@
...
 class EvervaultClient {
-  constructor(teamId) {
+  constructor(teamId, config) {
     if (!Datatypes.isString(teamId)) {
       throw new errors.InitializationError('teamId must be a string');
     }
-    this.config = Config(teamId);
+    if(!config.urls.keysUrl) {
+      if (!Datatypes.isString(config.urls.keysUrl)) {
+        config = undefined;
```
jakekgrog, line 12: If config is not defined correctly, shouldn't we fall back on default values? You're just setting it as undefined here

```suggestion
        config = DEFAULT_CONFIG;
```

lib/v1/index.js
```diff
@@ -1,14 +1,18 @@
 const { Datatypes, errors, extractDomain } = require('./utils');
 const Config = require('./config');
 const { Crypto, Http, Forms } = require('./core');
-const extractDomainFromUrl = require('./utils/domain');
 
 class EvervaultClient {
-  constructor(teamId) {
+  constructor(teamId, configOverride) {
```
lfarrel6, line 6:
```suggestion
  constructor(teamId, customConfig = {}) {
```

lib/v1/index.js
```diff
@@ -1,14 +1,18 @@
...
     if (!Datatypes.isString(teamId)) {
       throw new errors.InitializationError('teamId must be a string');
     }
-    this.config = Config(teamId);
+    if(!configOverride || !configOverride?.urls) {
+      configOverride = {
+        urls: undefined
+      }
+    }
```
lfarrel6, line 14:
```suggestion
    if(!customConfig) {
      customConfig = {};
    }
```

lfarrel6 (approved):

eoinboylan (approved):

PR #17, opened 2022-03-23
Update browser SDK with support for NIST-P256 curve

jakekgrog:
# Why
Update browser SDK with support for NIST-P256 curve

Tested encrypted outputs in a cage

# How


eoinboylan (approved):
Good job!

PR #18, opened 2022-03-24
Adding Input provision and handling

DeirdreCleary:
# Why
We need to provide iframes for collecting card details for customers' PCI requirements

# How
Added a function to generate an input which inserts the iframe into the client's page and returns a hook and method for getting the encrypted data
https://evervault.notion.site/evervault/Evervault-Inputs-59655cf4f0794faca8c255ccc73703b4

### DIFF
--- a/lib/v1/config.js
+++ b/lib/v1/config.js
@@ -1,5 +1,6 @@
 const KEYS_URL = 'https://keys.evervault.com';
 const METRICS_URL = 'https://metrics.evervault.com';
+const INPUTS_URL = 'https://inputs.deirdre.workers.dev/';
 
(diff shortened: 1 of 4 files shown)


lib/v1/config.js
```diff
@@ -1,5 +1,6 @@
 const KEYS_URL = 'https://keys.evervault.com';
 const METRICS_URL = 'https://metrics.evervault.com';
+const INPUTS_URL = 'https://inputs.deirdre.workers.dev/';
```
DeirdreCleary, line 3: This will need updating once we have the prod url

edmundojr (approved):

jakekgrog (approved):

PR #34, opened 2022-05-25
Use uncompressed P-256 key for Safari support

shanecurran:


lfarrel6:
Changes included in #35

PR #38, opened 2022-07-26
feat!: Add required appId in initialization

jakekgrog:
BREAKING CHANGE: `appId` is now required to during initialization. This will be used to fetch the correct api keys for cages for an app.

Introducing apps


lib/v2/index.js
```diff
@@ -0,0 +1,144 @@
...
+const Config = require("./config");
+const { Crypto, Http, Forms, Input } = require("./core");
+const { deriveSharedSecret } = require("../utils/crypto");
+
+class EvervaultClient {
+  constructor(teamId, appId, customConfig = {}) {
+    if (!Datatypes.isString(teamId)) {
+      throw new errors.InitializationError("teamId must be a string");
+    }
+    if (!Datatypes.isString(appId)) {
```
e-moran, line 11: Should we validate that app id looks correct? Not being able to do so for teamId is annoying because it is leading to the keys issue when the teamId is invalid

e-moran (commented):
Just one question, also need to update evervault-react to use this v2 and accept appId in the context provider.

e-moran (approved):

PR #45, opened 2023-01-05
feat: Ability to encrypt files in the Evervault Javascript SDK

e-moran:
# Why
Javascript SDK needs to add support for the encryption of Files in the browser.

# How
- Added logic to check if an object passed in to `evervault.encrypt()` is a file. If so it is encrypted per our File Encryption specification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,6 @@ The Evervault JavaScript SDK exposes two functions.
 
 `evervault.encrypt()` encrypts data for use in your [Cages](https://docs.evervault.com/tutorial). To encrypt data in the browser, simply pass an object or string into the `evervault.encrypt()` function. Store the encrypted data in your database as normal.
 
-Note: Encrypted Files or Blobs are returned as an ArrayBuffer containing the bytes of the encrypted File.
-
 ```javascript
 async evervault.encrypt(data: Object | Array | String | Number | File | Blob);
 ```
--- a/README.md
+++ b/README.md
@@ -34,8 +34,10 @@ The Evervault JavaScript SDK exposes two functions.
 
 `evervault.encrypt()` encrypts data for use in your [Cages](https://docs.evervault.com/tutorial). To encrypt data in the browser, simply pass an object or string into the `evervault.encrypt()` function. Store the encrypted data in your database as normal.
 
+Note: Encrypted Files or Blobs are returned as an ArrayBuffer containing the bytes of the encrypted File.
+
 ```javascript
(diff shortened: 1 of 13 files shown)


lib/utils/datatypes.js
```diff
@@ -12,6 +12,7 @@ const isDefined = (data) => typeof data !== "undefined" && data !== null;
 const isUndefined = (data) => typeof data === "undefined";
 const isNumber = (data) => typeof data === "number";
 const isBoolean = (data) => typeof data === "boolean";
+const isFile = (data) => data instanceof File || data instanceof Blob;
 
 const isEncryptable = (data) =>
   isDefined(data) && (isString(data) || isNumber(data) || isBoolean(data));
```
DeirdreCleary, line 18: Do we need `isFile` in here too? Noticed the `isFile` check in `crypto.js` is inside an `isEncryptable` check

e-moran: I decided to just remove it from the one in `crypto.js` because it would've ended up encrypting Files that are contained within Objects which is a bit of a strange behaviour.

lib/v2/index.js
```diff
@@ -35,7 +42,12 @@ class EvervaultClient {
   }
 
   async loadKeys() {
-    const cageKey = await this.http.getCageKey();
+    if (this._ecdhTeamKey) return;
+
+    console.log(`debug: ${this.isInDebugMode()}`);
```
davidnugent2425, line 47: Just for local debugging?

```suggestion
```

README.md
```diff
@@ -34,8 +34,10 @@ The Evervault JavaScript SDK exposes two functions.
 
 `evervault.encrypt()` encrypts data for use in your [Cages](https://docs.evervault.com/tutorial). To encrypt data in the browser, simply pass an object or string into the `evervault.encrypt()` function. Store the encrypted data in your database as normal.
 
+Note: Encrypted Files or Blobs are returned as an ArrayBuffer containing the bytes of the encrypted File.
```
davidnugent2425, line 37: Is it still the case that they're returned in a Buffer?

davidnugent2425 (approved):
🔥
We probs can't merge yet though, as the deployment is automated, and I assume we're gonna release everything together?

README.md
e-moran: Good spot, forgot to update it. Will fix now!

davidnugent2425 (approved):

PR #60, opened 2023-03-13
feat: allow supplying the public key

DeirdreCleary:
# Why
Customer request to support supplying the public key rather than calling out to the API for it.

# How
Allowed an additional param in the supplied config object to specify the `publicKey`. If present the call to the keys.evervault.com endpoint does not take place


lib/v1/index.js
```diff
@@ -24,6 +24,13 @@ class EvervaultClient {
     this.forms = Forms(this);
     this.forms.register();
     this.input = Input(this.config);
+
+    if (customConfig.publicKey) {
```
boilsquid, line 28: Can we add this as a test case please!

boilsquid (changes requested):

lib/v1/index.js
DeirdreCleary: Done ✅

DogPawHat (approved):

PR #78, opened 2023-03-22
[ETR-649] Add basic playwright tests

DogPawHat:
# Why
We need playwright tests to make sure this works in an actual browser

# How
Wrote a simple app and wrote the tests against that

### DIFF
--- a/e2e/basic.spec.ts
+++ b/e2e/basic.spec.ts
@@ -1,17 +1,20 @@
 import { test, expect } from "@playwright/test";
+import * as dotenv from "dotenv";
+
+dotenv.config();
 
 const encryptedStringRegex =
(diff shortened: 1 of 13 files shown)


e2e/basic.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  /((ev(:|%3A))(debug(:|%3A))?(([A-z0-9+/=%]+)(:|%3A))?((number|boolean|string)(:|%3A))?(([A-z0-9+/=%]+)(:|%3A)){3}(\$|%24))|(((eyJ[A-z0-9+=.]+){2})([\w]{8}(-[\w]{4}){3}-[\w]{12}))/;
+
+test("has title", async ({ page }) => {
+  await page.goto(
+    `http://localhost:3000/?team=${process.env.EV_TEAM_UUID}&app=${process.env.EV_APP_UUID}`
+  );
+
+  // Expect a title "to contain" a substring.
+  await expect(page).toHaveTitle(/Evervault Test/);
+});
```
thomascullen, line 16: This isn't really testing our stuff so I think we can remove it.

thomascullen (approved):
Nice! Think we can remove the title test but other than that looks good 👏

PR #89, opened 2023-03-30
[ETR-650] Replace Buffer Polyfill

DogPawHat:
# Why

Should not be relying on the node js polyfill for Buffer

# How
Replaced with the `uint8arrays` helper library as Buffer is a superclass of that anyway. The npm package `uint8arrays` is better maintained and browser compatible.

### DIFF
--- a/lib/core/crypto.js
+++ b/lib/core/crypto.js
@@ -1,6 +1,9 @@
 import { Datatypes, errors, cryptoUtils } from "../utils";
 import concatUint8Arrays from "../utils/concatUint8Arrays";
-import { uint8ArrayToBase64String, utf8StringToUint8Array } from "../utils/encoding";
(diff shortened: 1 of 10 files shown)


e-moran:
Could we use [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) instead of adding a new dependency?

The multiformats dependency that it pulls in is 256kb which kinda defeats the purpose of removing the Buffer polyfill.

DogPawHat:
Replaced `uint8arrays` with some basic functions from ChatGPT

Compared with the old Buffer polyfill:

### Buffer ployfill
image

### Vanilla/ChatGPT
image


About 55% small or more.

e-moran (approved):

PR #99, opened 2023-05-08
feat: add crc32 checksum to file encryption

e-moran:
# Why
File encryption format now includes a crc32 checksum.

# How
Added the checksum to the end of encrypted files.
Included a util for calculating crc32 so that we don't add extra dependencies.


DogPawHat:
Fix tests and add in a changeset as a minor release.

DogPawHat (approved):

PR #138, opened 2023-06-06
fix: fix onload event in firefox

e-moran:
# Why
Firefox does not support `window.onload`

# How
Using window.addEventListener instead


DogPawHat (commented):
add changeset

DogPawHat (approved):

PR #143, opened 2023-06-07
Use primary-color instead of label color when focused for material theme

thomascullen:
# Why

Requested by Stitch.

[Slack thread](https://evervault.slack.com/archives/C04BTBHRYCV/p1685443320523839)

# How
Use primary color for label when the input is focused.

### DIFF
--- a/.changeset/stupid-insects-occur.md
+++ b/.changeset/stupid-insects-occur.md
@@ -1,0 +1,5 @@
+---
+"@evervault/inputs": minor
+---
+
+Use primary color for focused input labels in material theme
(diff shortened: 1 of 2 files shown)


DogPawHat (changes requested):
This will need a changeset, and I also need an explaniation  as to how this won't be a breaking change.

thomascullen:
This wont need a new version release though right? Because it's just styling updates on the hosted inputs.

thomascullen:
Technically this will effect people already using the material theme, however, this is the approach that @nevflynn wanted to take rather than introducing further config variables which I think makes sense.

DogPawHat (approved):

PR #146, opened 2023-06-08
Fixes resolving promises when encrypting array

lammertw:
Make sure to wait for all promises in array to be resolved when encrypting an array

# Why
The traverseObject function is currently returning an array of promises instead of waiting for all promises to be resolved.

# How
Using await Promise.all on the array of promises waits for the promises to be resolved.


DogPawHat:
The github actions are acting weird here. I'm gonna clone this myself.

DogPawHat:
Now it's working. The Omnissiah works in mysterious ways.

PR #195, opened 2023-09-07
add new html and add copy

e-moran:
# Why
Adds more customisation to Reveal.
Adds a Copy button to Reveal
Fixed a bug where Inputs wouldn't encrypt in Debug mode for free teams

# How
Use a new HTML page for Reveal so that its completely separate from Inputs


packages/browser/README.md
```diff
@@ -185,6 +185,57 @@ inputs.isInputsLoaded.then(() => {
...
+Once you have your endpoint that returns the encrypted card data, you'll need to create an
+[Evervault Inbound Relay](https://docs.evervault.com/products/inbound-relay) that will decrypt the
+encrypted card data as is passes through it, before it gets to the iFrame. Once you have created
+your Relay you can add the component to your React app.
+
+```html
+<div id="reveal"></div>
+
+<script>
+  let request = new Request("https://secure.example.com", {
```
thomascullen, line 215: Should we change this to something like `https://my-endpoint.relay.evervault.com` to indicate it should be the relay endpoint?

packages/browser/lib/core/inputs.ts
```diff
@@ -38,6 +42,7 @@ export default function Inputs(config: Config) {
 
       const isInputsLoaded = new Promise<boolean>((resolve) => {
         window.addEventListener("message", (event) => {
+          console.log(event);
```
thomascullen, line 45: Remove `console.log`?

packages/browser/lib/utils/constructSource.ts
```diff
@@ -21,9 +24,14 @@ export default function constructSource(
...
+  let { inputsOrigin } = config.input;
   const { teamId, appId } = config;
 
+  let inputsUrl = inputsOrigin + INPUTS_URL_SLUG;
+  if (isReveal) {
+    inputsUrl = inputsOrigin + REVEAL_URL_SLUG;
+  }
+
   return `${inputsUrl}?team=${teamId}&app=${appId}${settingsQuery}&mode=${
     isReveal ? "reveal" : "inputs"
```
thomascullen, line 36: Do we still need to pass mode if we have a separate html file for reveal now?

thomascullen (changes requested):
Nice! 👏 Couple of small nits. Ping me when ready for review again and will do full QA.

packages/browser/README.md
e-moran: Great idea!

packages/browser/lib/core/inputs.ts
e-moran: Yes, nice catch!

thomascullen (approved):
Played around locally and all seems to work! Nice job 👏

PR #210, opened 2023-09-22
Prevent material theme from overriding .hide class

thomascullen:
# Why
The `material` theme was applying `display: flex` which was overriding the `.hide` class.

# How
Restrict the material theme CSS selector to  input groups that don't have the hide class.


e-moran (approved):
🚢

PR #224, opened 2023-11-09
Jake/etr 1356 embed data roles in ciphertext browser sdk

jakekgrog:
# Why
The `encrypt` function has been enhanced to accept an optional Data Role. This role, once specified, is associated with the data upon encryption. Data Roles can be created in the Evervault Dashboard (Data Roles section) and provide a mechanism for setting clear rules that dictate how and when data, tagged with that role, can be decrypted.

`encrypt(data, 'my-role')`


packages/browser/lib/core/crypto.ts
```diff
@@ -248,21 +337,71 @@ export class CoreCrypto {
...
+
+  async #traverseObject(
+    data: Datatypes.EncryptableValue,
+    role?: string
   ): Promise<Datatypes.EncryptedValue>;
 
-  async #traverseObject(data: unknown) {
+  async #traverseObject(data: unknown, role?: string) {
     if (Datatypes.isEncryptableAsString(data)) {
       return this.#encryptString(
```
thomascullen, line 406: Need to pass the role to the `encryptString` call here.

packages/browser/lib/core/crypto.ts
```diff
@@ -227,6 +295,27 @@ export class CoreCrypto {
...
+      metadataOffset[1] = (metadataBytes.length >> 8) & 0xff;
+      dataToEncrypt = concatUint8Arrays([
+        metadataOffset,
+        metadataBytes,
+        utf8StringToUint8Array(str),
+      ]);
+    } else {
+      version = this.#config.versions.NOC;
+      dataToEncrypt = utf8StringToUint8Array(str);
+    }
```
thomascullen, line 317: Do we only assign metadata if a role is present in the other SDK's? Given the `buildMetadata` function doesn't require a role I think it might make sense to always assign metadata to have the encrypted at timestamp.

thomascullen (changes requested):
One small issue when encrypted objects.

packages/browser/lib/core/crypto.ts
jakekgrog: We should be doing this everywhere but some SDKs dont. Not sure why/where that diverged but I'll add it here since we get more benefit out of it.

packages/browser/lib/core/crypto.ts
jakekgrog: good catch

thomascullen (approved):

PR #249, opened 2023-12-20
Support displaying card holder name field via fields argument

thomascullen:
# Why

Browsers do not allow input auto completes to pass from the parent frame down into the child iframe. This meant that when prompted on a 'name' field in the parent document, the auto complete would not fill out the card number, expiry or cvc.

# How

The only way to fix this at the moment is to provide functionality for users to render a 'name' field inside of the component iFrame. This PR adds a `fields` option to the Card component which allows users to define which fields they want to display and supports a 'name' field.

```js
const card = evervault.ui.card({
  fields: ['name', 'number', 'expiry', 'cvc'],
})
```

This PR also includes a new `isComplete` field on the card payload which indicates if the card has been correctly filled out. As well as a new `complete` event which will fire once the complete is filled out successfully.

### DIFF
--- a/.changeset/chilly-bees-lick.md
+++ b/.changeset/chilly-bees-lick.md
@@ -1,0 +1,7 @@
+---
+"@evervault/ui-components": minor
+"@evervault/browser": minor
+"@evervault/react": minor
+---
+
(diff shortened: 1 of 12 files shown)


jeremywagemans (approved):
🚢

PR #268, opened 2024-03-12
Update card number input masks

thomascullen:
# Why
Currently we apply the same 19 digit mask to card number inputs regardless of the brand of the card. Only Union pay cards support 19 digit card numbers. Similarly Amex cards only have 15 digits and have a different formatting of `0000 0000000 00000`.

# How
Update both UI Components and Inputs to make the card number mask dependent on the type of the card as the user enters it.

- The default mask is now 16 digits
- Unionpay cards support up to 19 digits
- Amex cards are formatted appropriately and only support 15 digits.

**Note: This includes a major version bump for the IMask dependency in the Inputs package.**


thomascullen:
@SocketSecurity ignore-all

boilsquid (approved):

PR #281, opened 2024-04-23
[ETR-1862] Add ability to restrict accepted card brands

jeremywagemans:
# Why
We are adding the ability to restrict accepted brands. A merchant only accepting Visa & Mastercard cards can now restrict this using the `acceptedBrands` field. When a card is valid but its brand is not accepted, the card is marked as invalid.

We're also fixing a bug where an 8-digit bin is returned when the card number length is < 16 digits.

# How
- A new `localBrand` field is added to the card object. This field is used to expose all local brands a card has been issued with
- A new `acceptedBrand` option is added to the card function to restrict the list of accepted brands.
- `bin` contains 6 digits when the card number is shorter than 16 digits


thomascullen (changes requested):
Couple of small changes needed but nothing major. Nice work!

thomascullen (approved):
All looks good to me! Nice one.

PR #282, opened 2024-04-25
update with payload

boilsquid:
# Why
The current implementation breaks form validation for HTML fields that have a required attribute. This is because the js creates shadow copies of the DOM elements and submits them with encrypted values instead of the real ones, causing a validation error on the empty original elements.

# How
- Attach an event handler to the detected form and remove its action attribute, this stops the iframe submit the form. Then cache the action value and attach an `onClick` handler to the form. When the submit button is clicked halt the event, encrypted the required field values, reset the form action and submit the form.

### DIFF
--- a/packages/browser/lib/main.ts
+++ b/packages/browser/lib/main.ts
@@ -360,39 +360,52 @@ export default class EvervaultClient {
       );
     } else {
       forms.forEach((form: Form) => {
-        const hiddenInput = findFormByHiddenField(form.uuid);
+        const hiddenInput = findFormByHiddenField(
+          form.uuid,
+        );
         if (hiddenInput === null) {
           return;
         }
+
         const parentForm = findParentOfInput(hiddenInput);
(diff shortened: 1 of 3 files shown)


packages/browser/lib/main.ts
```diff
@@ -364,35 +364,46 @@ export default class EvervaultClient {
...
+                  });
+              })
+            );
           });
+          Promise.all(mutations)
+            .then(() => {
+              parentForm.setAttribute("action", actionAttr);
+              parentForm.submit();
+            })
+            .catch(() => console.error("Error encrypting form value"));
```
hanneary, line 406: any point in logging the error here for context?

hanneary (approved):

packages/browser/lib/main.ts
boilsquid: Sure!

boilsquid: Sure

PR #299, opened 2024-05-23
Add React Native SDK

doylemark:
# Why
This imports the rn sdk directory to packages and makes it work with the pnpm workspace.

The WIP components and any other functionality of the RN sdk can be ignored in this PR

# How


packages/card-validator/index.ts
doylemark, line 1: All changes here are to stop the TS compiler complaining when this is imported in the rn lib

packages/card-validator/package.json
doylemark, line 1: having the exports/repo/files section seems to break locally import from the pnpm workspace, we don't need to publish this now that the rn lib is in the workspace anyway

boilsquid (approved):

PR #305, opened 2024-05-24
Release new card validators version

doylemark:
# Why
changeset didn't apply previously and release never happened. We actually want to publish this again now that we can't use shared packages in react native

# How
add back the distribution stuff to package.json

### DIFF
--- a/packages/card-validator/package.json
+++ b/packages/card-validator/package.json
@@ -1,8 +1,30 @@
 {
   "name": "@evervault/card-validator",
-  "version": "1.0.1",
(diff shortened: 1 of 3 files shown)


packages/card-validator/package.json
```diff
@@ -1,8 +1,30 @@
 {
   "name": "@evervault/card-validator",
-  "version": "1.0.1",
+  "version": "1.0.2",
```
doylemark, line 3: are we supposed to be bumping this by hand or is changeset meant to do it?

PR #345, opened 2024-07-01
Validate CVC when no card number present

DeirdreCleary:
# Why
A `Card` could be used to collect only CVC, in which case the type of the card should not be considered

# How
When no card number is present, only check that CVC is a 3 or 4 digit number

### DIFF
--- a/packages/card-validator/index.ts
+++ b/packages/card-validator/index.ts
@@ -107,16 +107,23 @@ export function validateCVC(
   cvc: string,
   cardNumber: string
 ): CardCVCValidationResult {
-  const validatedCard = validateNumber(cardNumber);
-  if (!validatedCard.isValid) {
+  // Check if the CVC only contains numbers with 3 or 4 digits
+  if (!/^\d{3,4}$/.test(cvc)) {
     return {
       cvc: null,
       isValid: false,
     };
   }
 
-  // Check if the CVC only contains numbers
-  if (!/^\d*$/.test(cvc)) {
+  if (!cardNumber) {
+    return {
+      cvc: cvc,
+      isValid: true,
+    };
+  }
+
+  const validatedCard = validateNumber(cardNumber);
+  if (!validatedCard.isValid) {
     return {
       cvc: null,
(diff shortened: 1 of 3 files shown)


packages/card-validator/index.ts
```diff
@@ -107,16 +107,23 @@ export function validateCVC(
   cvc: string,
   cardNumber: string
 ): CardCVCValidationResult {
-  const validatedCard = validateNumber(cardNumber);
-  if (!validatedCard.isValid) {
+  // Check if the CVC only contains numbers with 3 or 4 digits
```
jeremywagemans, line 110: I don't think we should be doing this because it will then always allow 3-4 digits even when we know the brand (which is the majority of cases). I'd probably make cardNumber optional and fallback on the regex when cardNumber is null/undefined?

packages/card-validator/index.ts
```diff
@@ -107,16 +107,23 @@ export function validateCVC(
   cvc: string,
   cardNumber: string
```
thomascullen, line 108: We need to mark the cardNumber arg as optional now I think

thomascullen (commented):
Nice, fast turn around! One tiny change with card validator arg types and we should be good to go

packages/card-validator/index.ts
DeirdreCleary: Done, and tests added 👍 Thanks!

packages/card-validator/index.ts
```diff
@@ -107,16 +107,23 @@ export function validateCVC(
...
 ): CardCVCValidationResult {
-  const validatedCard = validateNumber(cardNumber);
-  if (!validatedCard.isValid) {
+  // Check if the CVC only contains numbers with 3 or 4 digits
+  if (!/^\d{3,4}$/.test(cvc)) {
```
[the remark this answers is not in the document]
boilsquid: Sorry very good, didnt see that thread!

thomascullen (approved):
👏

PR #361, opened 2024-07-30
Remove spaces from CC Num before encryption

doylemark:
# Why
Causing issues with payment processors

# How
Remove spaces before encryption occurs


thomascullen (approved):
Need to

PR #363, opened 2024-07-30
Add support for 3DS Method

thomascullen:
Adds [3DS method](https://docs.3dsecure.io/3dsv2/3dsmethod.html) support to the ThreeDSecure UI Component


jeremywagemans (approved):
Looks great 🙏

PR #376, opened 2024-08-16
Add validate event

thomascullen:
The `validate` method on the card component can be used to manually trigger validation to display any error messages inside of the component. Due to everything having to be done via messaging in the iframe this is an async call and does not immediately provide results when called. The `validate` method was not intended to provide results but instead to simply cause any error messages to be shown, when the user hasn't interacted with the inputs. e.g Submits without filling out any details. 

It could be useful however, to provide some way for users to run logic after validation has occurred. In an ideal world, we would simply make this method return a promise which can be awaited, however, unfortunately that would be a breaking change and so isn't an option. This PR adds a new `validate` event which is fired after validation has completed and is passed the latest form state.

This also updates the `card.values` attribute to have a default empty values object instead of `undefined`

### DIFF
--- a/examples/collect-card-details/src/main.ts
+++ b/examples/collect-card-details/src/main.ts
@@ -24,6 +24,10 @@ card.on("swipe", (values) => {
   console.log("swipe", values);
 });
 
+card.on("validate", (values) => {
+  console.log("validate", values);
+});
+
 card.mount("#form");
 
 const btn = document.getElementById("purchase");
--- a/examples/collect-card-details/src/main.ts
+++ b/examples/collect-card-details/src/main.ts
@@ -38,7 +38,7 @@ btn?.addEventListener("click", () => {
   output.innerText = "";
   card.validate();
 
-  if (card.values?.isValid) {
+  if (card.values.isComplete) {
     console.log("Valid!", card.values);
     const { number, expiry, cvc } = card.values.card;
     output.innerHTML += "Thank you for your purchase! <br /><br />";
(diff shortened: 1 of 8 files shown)


examples/collect-card-details/src/main.ts
```diff
@@ -24,6 +24,10 @@ card.on("swipe", (values) => {
   console.log("swipe", values);
 });
 
+card.on("validate", (values) => {
```
doylemark, line 27: type annotation on values could be nice

packages/shared/src/useForm.ts
```diff
@@ -163,9 +171,15 @@ export function useForm<T extends object>({
...
-    triggerChange.current = false;
-    onChange?.(form);
+    if (triggerChange.current) {
+      triggerChange.current = false;
+      onChange?.(form);
+    }
+
+    if (validationCallback.current) {
+      validationCallback.current(form);
+      validationCallback.current = undefined;
```
doylemark, line 181: why do we set the ref to undefined here? These should end up stale anyway since they're not in the dep array

thomascullen: Refs are outside of the render cycle so they have no effect on the dependency array. ( as far as I know 😅 )

doylemark: TIL.. Shame that eslint hasn't been telling me this for the last 5 years 😔

doylemark (approved):
lgtm!

PR #379, opened 2024-08-23
Add a new provider to ease initialisation of RN SDK

doylemark:
# Why
At the moment it's a little bit awkward to setup the RN sdk because you need to call an `init` method inside an effect.
This isn't very idiomatic for React.

# How
Create an `EvervaultProvider` that initialises the SDK. Althought this isn't actually a context provider at the moment, I imagine it will become one in the future.


donaltuohy (approved):

packages/react-native/src/components/EvervaultProvider.tsx
```diff
@@ -0,0 +1,65 @@
...
+  appUuid?: string;
+}
+
+const Ctx = createContext<EvervaultContext>({
+  teamUuid: undefined,
+  appUuid: undefined,
+});
+
+// this isn't publicly exposed
+export const useEvervault = () => useContext(Ctx);
```
doylemark, line 17: we want this for 3ds etc

thomascullen (approved):

PR #382, opened 2024-08-26
Use teamId instead of teamUuid in public facing props

doylemark:
# Why
Better UX

# How
Update props

### DIFF
--- a/packages/react-native/src/components/EvervaultProvider.tsx
+++ b/packages/react-native/src/components/EvervaultProvider.tsx
@@ -1,6 +1,5 @@
 import { ReactNode, createContext, useContext, useEffect } from "react";
 import * as React from "react";
-import { init } from "../sdk";
 import { EvervaultSdk } from "../native";
 
 export interface EvervaultContext {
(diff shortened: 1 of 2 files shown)


donaltuohy (approved):

packages/react-native/src/components/EvervaultProvider.tsx
```diff
@@ -36,30 +35,34 @@ interface EvervaultProps {
  * }
  * ```
  */
-const EvervaultProvider = ({ teamUuid, appUuid, children }: EvervaultProps) => {
+const EvervaultProvider = ({ teamId, appId, children }: EvervaultProps) => {
   useEffect(() => {
     async function initEvervault() {
       try {
-        if (!teamUuid?.startsWith("team_")) {
+        if (!teamId?.startsWith("team_")) {
```
donaltuohy, line 42: Actually just thinking, some of the older teams don't have this prefix so they wouldn't be able to use this

doylemark: good point will remove :(

PR #386, opened 2024-08-29
[ETR-2493] add 3ds to react native sdk

donaltuohy:
# Why
Add 3DS to react native SDK. 

Docs PR here: https://github.com/evervault/docs/pull/282

TODO: 
 - [x] Update webview URL once jspackage is live

# How
This PR introduces three new exports to the SDK, providing two ways to trigger 3DS:

1. `<ThreeDSecure.Modal />` with `useThreeDSecure()` - Out-of-the-Box Solution

**IOS**

https://github.com/user-attachments/assets/f0af9f2b-27d4-4262-a6af-50d2621c8bf3

**Android**

https://github.com/user-attachments/assets/237d6e81-4962-4430-839c-bbd999e16395

The ThreeDSecure.Modal component provides a fully managed 3DS flow that handles everything under the hood. It displays a modal with the 3DS webview when required. This component simplifies the 3DS process and allows for custom styling and configuration through props.

```js

function Checkout({ cardData }: { cardData: CardPayload | undefined }) {
  const tds: UseThreeDSecureResponse = useThreeDSecure();

  const handlePayment = async () => {
    const sessionId = await create3DSecureSession({
      cardNumber: cardData.card.number,
      expiryMonth: cardData.card.expiry.month,
      expiryYear: cardData.card.expiry.year,
    });

    tds.start(sessionId, {
      onSuccess: () => {
        console.log("3DS successful");
      },
      onFailure: (error: Error) => {
        console.error("3DS failed", error);
      },
      onError: (error: Error) => {
        console.log("3DS errored", error);
      },
    });
  };

  return (
    <>
        <ThreeDSecure state={tds} >
               <ThreeDSecure.Modal />
        <ThreeDSecure />
        <Button title="Pay" onPress={handlePayment} />
    </>
  );
}

export default function App() {
return (
  <EvervaultProvider
      teamId={process.env.EXPO_PUBLIC_EV_TEAM_UUID}
      appId={process.env.EXPO_PUBLIC_EV_APP_UUID}
    >
         <Checkout />
    </EvervaultProvider>
})
```

2 `ThreeDSecure` - Custom Implementation Provider

The `ThreeDSecure` component allows for more customization. It acts as a provider that takes in a sessionState. This provider can wrap custom components, including the ThreeDSecure.Frame for the webview.

```js
function CustomThreeDSecureCheckout({cardData}: {cardData: CardPayload | undefined}) {
  const tds = useThreeDSecure();

  const handlePayment = async () => {
    const sessionId = await create3DSecureSession({
      cardNumber: cardData?.card.number,
      expiryMonth: cardData?.card.expiry.month,
      expiryYear: cardData?.card.expiry.year,
    });

    tds.start(sessionId, {
      onSuccess: () => {
        console.log("3DS successful");
      },
      onFailure: (error: Error) => {
        console.error("3DS failed", error);
      },
      onError: (error: Error) => {
        console.log("3DS errored", error);
      },
    });
  }

  const closeCustomModal = async () => {
    await tds.cancel();
  };

return (
    <ThreeDSecure state={tds}>
        <View style={styles.modalContainer}>
          <View style={styles.modalContent}>
            <View style={styles.modalHeader}>
            <TouchableOpacity onPress={closeCustomModal}>
              <Text>CLOSE MODAL</Text>
            </TouchableOpacity>
            </View>
            <ThreeDSecure.Frame />
          </View>
        </View>
    </ThreeDSecure>
 )
}
```


packages/react-native-example/App.tsx
```diff
@@ -24,24 +32,108 @@ if (
...
+      console.error("3DS failed", error);
+    },
+    onError: (error: Error) => {
+      console.log("3DS errored", error);
+    },
+  });
+
+  const handlePayment = () => {
+    session.start();
+  }
```
thomascullen, line 51: Instead of passing the session id as the first param of `useThreeDSecure` we should pass it as an argument to `start`. The reason for this is that users will need to make a backend call to create the 3DS session before they render the component and it's a little neater to be able to do all of this as part of a single submit handler.

e.g

```js
const tds = useThreeDSecure()

async function handleSubmit() {
  const session = await createSession()
  tds.start(session.id)
}
```

thomascullen (commented):
This is great, thanks for putting all of this together! Have left a few small comments.

One slightly larger thing i'm wondering is if we should change the approach slightly to always use the `useThreeDSecure` hook and the `ThreeDSecure` provider whether you are using custom or modal approach. This would remove the need for the useCancel hook and might align the two flows a little more removing the need to pass callback objects into the components.

e.g

```js
// modal approach
function Checkout() {
  const tds = useThreeDSecure()

  const checkout = () => {
    // call to backend to create 3ds session
    tds.start("sesion_id")
  }

  return (
    <>
      <button onClick={checkout}>Checkout</button>
      <ThreeDSecure state={tds}>
        <ThreeDSecure.Modal />
      </ThreeDSecure>
    </>
}

// custom approach
function Checkout() {
  const tds = useThreeDSecure()

  const checkout = () => {
    // call to backend to create 3ds session
    tds.start("sesion_id")
  }

  return (
    <>
      <button onClick={checkout}>Checkout</button>
      <ThreeDSecure state={tds}>
        <View style={styles.modalContainer}>
          <View style={styles.modalContent}>
            <TouchableOpacity onPress={tds.cancel}>
              <Text>CLOSE</Text>
            </TouchableOpacity>
            <ThreeDSecure.Frame />
          </View>
        </View>
      </ThreeDSecure>
    </>
}
```

In these cases instead of the `shouldDisplayModal` value in the `ThreeDSecure` provider component we could store the session id in state in the hook and return null if that isn't set. Users would use the start method to set it.

```jsx
function ThreeDSecure({ state, children }) {
  ...

  if (state.session) return children
  return null
}
```

packages/react-native/src/components/3DS/ThreeDSecureFrame.tsx
```diff
@@ -0,0 +1,39 @@
...
+import WebView from "react-native-webview";
+import { CHALLENGE_DOMAIN_3DS } from "./config";
+import { Spinner } from "./Spinner";
+import { defaultThreeDSecureFrameStyles as defaultStyles } from "./styles";
+
+interface ThreeDSecureFrameProps {
+  sessionId: string;
+}
+
+export function ThreeDSecureFrame({ sessionId }: ThreeDSecureFrameProps) {
```
thomascullen, line 13: Rather than taking the sessionId as a prop it would be cleaner to pull it from the context of the `ThreeDSecure` provider.

```
const { sessionId } = useContenxt(ThreeDSecureContext)
```

packages/react-native/src/components/3DS/useThreeDSecure.tsx
```diff
@@ -0,0 +1,63 @@
...
+import {
+  ThreeDSecureCallbacks,
+  UseThreeDSecureResponse,
+  UseThreeDSecureState,
+} from "./types";
+
+export const useThreeDSecure = (): UseThreeDSecureResponse => {
+  const { appUuid } = useEvervault();
+  const intervalRef = useRef<NodeJS.Timeout | null>(null);
+  const [displayModal, setDisplayModal] = useState(false);
```
thomascullen, line 14: Very nit picking but could we rename this var to `isOpen` or `isVisible` as it's now not strictly for the built in modal version.

packages/react-native/src/components/3DS/useThreeDSecure.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const [state, setState] = useState<UseThreeDSecureState>({
+    session: null,
+    displayModal,
+  });
```
thomascullen, line 18: No need to duplicate the modal state here. Instead we can just have.

```
const [session, setSession] = useState(null)
const [isVisible, setIsVisible] = useState(flase);
```

And then call the setSession and setIsVisible methods individually.

The return from the hook would become

```
return {
    start,
    cancel,
    session,
    isVisible
};
```

packages/react-native/src/components/3DS/useThreeDSecure.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  useEffect(() => {
+    setState((prevState) => ({
+      ...prevState,
+      displayModal,
+    }));
+  }, [displayModal]);
```
thomascullen, line 25: This can be removed if we separate the state like described in my previous comment.

packages/react-native/src/components/3DS/useThreeDSecure.tsx
```diff
@@ -0,0 +1,51 @@
...
+    } else {
+      console.warn("No 3DS session to cancel");
+    }
+  };
+
+  return {
+    start,
+    cancel,
+    session,
+    displayModal: isVisible,
```
thomascullen, line 49: Sorry I meant to rename across the board so that even uses of state would be `state.isVisible` inside of `ThreeDSecure` provider. 🙈

packages/react-native/src/components/3DS/ThreeDSecureModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+    throw new Error("ThreeDSecure.Modal must be used within an Evervault ThreeDSecure provider component");
+  }
+
+  const close3DS = async () => {
+    if (state?.session) {
+      await state.session.cancel();
+    };
+  };
+
+  if (state.session && state.displayModal) {
```
thomascullen, line 22: We should be able to remove the check for `state.displayModal` now that this will always be a child of the provider which has it's own check for that .

thomascullen (commented):
Two tiny suggestions!

packages/react-native/src/components/3DS/ThreeDSecureModal.tsx
donaltuohy: Good shout, updated!

thomascullen (approved):
a-man-with-a-mustache-is-saying-it-s-so-it-s-so-beautiful

PR #389, opened 2024-09-04
Resolve issue in S3 publishing

ghost:
# Why

Inputs and browser SDK where failing to publish due to incompatibility in the download and upload actions.

# How

bumped the upload action to v4


.github/workflows/deploy-ui-components.yml
```diff
@@ -40,32 +28,19 @@ jobs:
     runs-on: ubuntu-latest
     environment: ${{ inputs.environment }}
     name: ${{ inputs.name }}
-    env:
-      VITE_EVERVAULT_JS_URL: ${{ inputs.vite-evervault-js-url }}
-      VITE_KEYS_URL: ${{ inputs.vite-keys-url }}
-      VITE_API_URL: ${{ inputs.vite-api-url }}
```
thomascullen, line 46: These need to be set when building ui components so we will need to be sure to add these to the environment vars in action settings.

thomascullen (approved):
We need to add some environment variables in settings before merging.

PR #429, opened 2024-09-23
Fix: Support partial initial values for card form

lfarrel6:
# Why

The react native card component supports building a partial form, but the types require a complete card object for initial values. 

Also, the 3ds example implementation required a backend to send authenticated requests to the 3ds API.

# How

- Update the Card component's types to support a partial object being passed in.
- Update example RN app to have an example backend which calls the Evervault API with an API key for 3ds sessions
  - The structure of this isn't ideal, but the alternative of sending authenticated requests from the client isn't something that should ever happen.
- Format the example RN app


lfarrel6:
SocketSecurity ignore-all

lfarrel6:
@SocketSecurity ignore-all

PR #431, opened 2024-09-24
Update react native sdk to invalidate previously valid cvcs when card number changes

lfarrel6:
# Why

The CVC length is dependent on the card brand entered, so a previously valid CVC needs to be invalidated in the event that a card number changes its length requirement.

# How

- Update react native to revalidate the CVC every time the number changes, regardless of its current validity.
- Update e2e tests for react to account for updated behaviour.


lfarrel6:
@SocketSecurity ignore-all Package is included through axios which is not included in any public packages.

PR #454, opened 2024-10-04
Rewrite browser e2e tests

thomascullen:
# Why

Our e2e tests for the browser SDK rely on using a function to decrypt values and make assertions. This was put together before we had the decrypt API. This makes moving the tests to another team and app a little difficult as the app needs to deploy the function first.

# How
- Rewrite the tests to rely on the decrypt API instead of the function.
- Add more e2e tests
- Remove the two additional e2e-apps packages that were required to run the browser tests.
- Remove the function code as it's not longer needed.


boilsquid (approved):

thomascullen:
@SocketSecurity ignore-all

PR #463, opened 2024-10-08
Upgrade ESLint to version 9

thomascullen:
This uses community defaults and adds in new linters that we didn't have before such as linting for React hook rules. I've commented out any areas that have new offending rules for now to avoid any actual logic changes to packages.


thomascullen:
@SocketSecurity ignore-all

PR #469, opened 2024-11-06
Add events for Card component field interactions 

thomascullen:
Some users have requested functionality to track interactions with individual fields inside of the Card component. This PR introduces 4 new events for the Card component. `focus`, `blur`, `keyup` and `keydown`. These events hook into the corresponding native events for each input, however, it's important that we don't provide the original event which would allow users to gain access to plain text data. Instead we pass an object which includes the field the event originated from, as well as the same card data payload that is sent with the `change` event.

e.g 

```
{
  field: "number",
  data: {
    card: {
      name: "",
      brand: "visa",
      localBrands: [],
      bin: "42424242",
      lastFour: "4242",
      number: "ev:QkTC:WXkiIR/el/Sihcpd:AwBL9S7F91yMdpl2kKYY8tgEJJWGQbiyOX4ciF+arpdO:Q96nNPbHVNenK1NkAAIhaElyfL3u0IDyjrQhjFNPzZxvzVkCbpc7n4e8AkA96to:$",
      expiry: {
        month: "12",
        year: "25"
      },
      cvc: null
    },
    isValid: true,
    isComplete: false,
    errors: null
  }
}
``` 

Users requesting this functionality have also stated that they would like to be able to distinguish between the user entering the expiry month vs the expiry year. Currently, we only return the expiry values once the user has entered a complete expiry date. This PR also changes the expiry validation so that the parsed month and year is returned even when the entered value is invalid.

e.g 

"03" would return `{ month: "03", year: null, isValid: false }`
"0320" would return `{ month: "03", year: "20", isValid: false }`

### DIFF
--- a/.changeset/wet-peaches-cry.md
+++ b/.changeset/wet-peaches-cry.md
@@ -1,0 +1,12 @@
+---
+"example-ui-components": minor
+"@evervault/card-validator": minor
+"@evervault/ui-components": minor
+"@evervault/browser": minor
+"shared": minor
+"@evervault/react": minor
+"types": minor
+---
+
+- Adds focus, blur, keyup and keydown events to the Card component to track interactions with inputs inside of the Card component.
+- Updates the payload from the card component to include the parsed month & year even when the entered expiry is invalid. Previously the expiry value would only be returned when the entered value is valid.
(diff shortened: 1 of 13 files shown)


donaltuohy (approved):
Lgtm, tested out and works as expected!

PR #470, opened 2024-11-11
Unmount 3ds instance on error

dave-kav:
# Why

Context: https://evervault.slack.com/archives/C05TBHREAG7/p1731332793569229

# How

Unmount the 3DS component on error/failure

## Before
Modal renders, errors, doesn't re-render, no subsequent patch calls
before

## After
Modal renders, errors, re-renders, subsequent patch calls
after

### DIFF
--- a/packages/react/lib/ui/useThreeDSecure.ts
+++ b/packages/react/lib/ui/useThreeDSecure.ts
@@ -1,5 +1,5 @@
 import type Evervault from "@evervault/browser";
-import { useCallback, useRef } from "react";
+import { useCallback, useEffect, useRef } from "react";
 import { ComponentError, ThemeDefinition } from "types";
 import { useEvervault } from "../useEvervault";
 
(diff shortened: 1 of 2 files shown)


packages/react/lib/ui/useThreeDSecure.ts
```diff
@@ -23,37 +23,56 @@ export function useThreeDSecure(opts?: UseThreeDSecureOptions) {
 
   const start = useCallback(
     (session: string, callbacks?: UseThreeDSecureCallbacks) => {
-      if (instance.current) return;
+      if (instance.current) {
+        instance.current.unmount();
+        instance.current = null;
+      }
```
lfarrel6, line 29: just to clarify — successive calls to start a 3ds session will unmount and remount with new sessions?

packages/react/lib/ui/useThreeDSecure.ts
```diff
@@ -23,37 +23,56 @@ export function useThreeDSecure(opts?: UseThreeDSecureOptions) {
...
-          instance.current.on("error", callbacks.onError);
-        }
+          if (callbacks?.onError) {
+            instance.current.on("error", callbacks.onError);
+          }
 
-        instance.current.mount();
+          instance.current.mount();
+        } catch (error) {
+          instance.current = null;
```
lfarrel6, line 55: unmount needed here? if not, should we leave the session in current so a user can restart by manually unmounting/invoking start again?

lfarrel6 (commented):
couple clarifying questions

packages/react/lib/ui/useThreeDSecure.ts
dave-kav: Effectively yeah because the session is most likely in a failed or error state if the user feels a need to submit again. However, this does open up to multiple sessions being created is someone clicks rapidly, so we'll need to do something smarter

packages/react/lib/ui/useThreeDSecure.ts
dave-kav: I think not needed yeah

lfarrel6 (approved):

PR #480, opened 2024-11-22
Add wallet API

donaltuohy:
# Why
Add support for Google and Apple Wallet APIs.

# How
* Add two new UI components.
* Add browser methods to initialise them
* Update examples to use the new components


packages/ui-components/src/ApplePay/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+    if (buttonElement) {
+      buttonElement.addEventListener('click', handleClick);
+    }
+
+    return () => {
+      if (buttonElement) {
+        buttonElement.removeEventListener('click', handleClick);
+      }
+    };
+  }, []);
```
thomascullen, line 154: The warning we are getting here is because the handleClick is defined outside the scope of the useEffect callback. We should probably move it inside before `const buttonElement = button.current` and also will  need to add `app` and `config` as dependencies of the useEffect array on line #154.

packages/ui-components/src/GooglePay/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+        console.error(err);
+      }
+    }
+
+    const script = document.createElement("script");
+    script.src = "https://pay.google.com/gp/p/js/pay.js";
+    script.async = true;
+    script.onload = onLoad;
+    document.body.appendChild(script);
+  }, [app, config, send]);
```
thomascullen, line 105: Need to add `on` in the dependencies array here.

examples/react-google-wallet/api.js
```diff
@@ -0,0 +1,28 @@
...
+    let body = "";
+    request.on("data", (chunk) => {
+      body += chunk.toString();
+    });
+    request.on("end", () => {
+      resolve(JSON.parse(body));
+    });
+    request.on("error", reject);
+  });
+}
```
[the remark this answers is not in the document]
donaltuohy: Removed the file!

examples/react-google-wallet/src/App.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+      const inst = evervault.ui.googlePay(transaction, {
+        type: "pay",
+        color: "white",
+        borderRadius: 15,
+        size: { width: 400, height: "60px" },
+        allowedAuthMethods: ["PAN_ONLY", "CRYPTOGRAM_3DS"],
+        allowedCardNetworks: ["VISA", "MASTERCARD"],
+        process: async (data, {fail}) => {
+          console.log("Sending encrypted data to merchant", data);
```
shanecurran, line 38:
```suggestion

```

thomascullen (approved):

PR #495, opened 2024-12-11
Test coverage

thomascullen:
Adds test coverage metrics that will fail when they fall below a set threshold. For now I have set the threshold based on our current coverage, however, we aim to increase all metrics to at least 90% over time.

Due to the complexity in our setup ( we have to serve the JS SDK over the network as well as requiring UI Components to be running ) this unfortunately isn't straight forward. I have created a new package which extends the playwright test utils to collect metrics from the current page as well as any UI Component iFrames that appear throughout the test. This requires both the browser SDK and UI Components package to be built with [istanbul](https://istanbul.js.org/) instrumentation. I have update the vite config to do this as long as the `VITE_TEST_COVERAGE` env var is set to true.

> [!NOTE]
> This appears to increase the e2e test step time by 3m to ~11m. I'm assuming this is just due to the fact that we need to write a lot of coverage files to disk. Hopefully there will be some things we can do to speed them up.t


butlerx (approved):

lfarrel6 (approved):

thomascullen:
@SocketSecurity ignore-all

PR #524, opened 2025-01-08
Replace Wallet API merchant object with merchantId field

dave-kav:
# Why

Improve the DX of the Wallet API.

# How

Get the required merchant details from the new merchant frontend API.


thomascullen (approved):
nice-wow-wow-wee-wow-gif

PR #525, opened 2025-01-08
[ETR-2882] update apple pay frontend to use payment request api

donaltuohy:
# Why
Disbursements aren't supported by the Apple Pay JS API. We need to migrate away from it to the payment request api.

# How
Update apple pay component to use the apple pay request api.


examples/react-google-wallet/src/App.tsx
```diff
@@ -19,7 +19,7 @@ function App() {
         amount: 4300,
         currency: "USD",
         country: "US",
-        merchantId: "merchant_e930d3f7bf37",
+        merchant: "merchant_e930d3f7bf37",
```
dave-kav, line 22: You had it right the other way 😅

donaltuohy: I'm a mess 😂

PR #533, opened 2025-01-16
Add apple pay disbursements API

donaltuohy:
# Why
Apple pay disbursements API is needed.

# How
Now that we have switched to using the payment request API, add support for the disbursements API

Transaction now has a `type` field which can be `payment`(default) or `disbursement`. Apple pay accepts a disbursment transaction which allows merchants to create a transaction which sends money to the user's card.

Adyen and Stripe's disbursment API require card holder information so if the merchant needs this info, they can specify requiredRecipientDetials in the disbursement transaction. Like so:

```js
      const disbursementTransaction = evervault.transactions.create({
        amount: 4300,
        currency: "USD",
        country: "US",
        merchantId: "merchant_abc",
        requiredRecipientDetails: ["email", "phone", "name", "address"],
        type: "disbursement",
      }); 

      const apple = evervault.ui.applePay(disbursementTransaction, {
        type: "contribute",
        style: "white-outline",
        locale: "en-US",
        size: { width: "100%", height: "80px" },
        borderRadius: 10,
        allowedCardNetworks: ["visa", "masterCard"],
        process: async (data, {}) => {
          console.log("Sending encrypted data to merchant", data);

          await new Promise((resolve) => {
            setTimeout(resolve, 2000);
          });
        },
      });
```

This will add the cardholders contact details to the payload returned.

I've also added two optional override objects which allow the user to pass apple specific objects in to override the apple pay implementation. We can provide docs on what fields must be set if they want to use a feature which we don't yet support (eg: shipping, coupons):

```js
 const apple = evervault.ui.applePay(disbursementTransaction, {
    ...,
    paymentOverrides?: {
        paymentMethodData?: PaymentMethodData[];
        paymentDetails?: PaymentDetailsInit;
    };
    disbursementOverrides?: {
        disbursementDetails?: PaymentDetailsInit;
    };
```

There is also a bugfix which now takes the domain from the browser SDK and passes it to UI components. It was been taken directly from within UI components which was returning the UI components domain.


packages/types/uiComponents.ts
```diff
@@ -468,14 +481,47 @@ export interface TransactionLineItem {
...
+  requiredRecipientDetails?: RequiredRecipientDetail[];
+}
+
+export type TransactionDetails =
+  | PaymentTransactionDetails
+  | DisbursementTransactionDetails;
+
+export type TransactionDetailsWithDomain = TransactionDetails & {
+  domain: string;
+};
```
dave-kav, line 518: Do we have any current usages of `TransactionDetails` on its own now? Seems we only use `TransactionDetailsWithDomain`, and no other direct usages of `TransactionDetails`? I'm just wondering out-loud whether it would make sense to add `domain` on to `BaseTransactionDetails` and then use `TransactionDetails` everywhere.

packages/types/uiComponents.ts
```diff
@@ -468,14 +481,47 @@ export interface TransactionLineItem {
...
+
+export type CreateTransactionDetails = Omit<
+  PaymentTransactionDetails,
+  "type"
+> & { type?: "payment" };
```
dave-kav, line 523: How do you feel about
```ts
export type TransactionDetailsWithDomain = TransactionDetails & {
  domain: string;
};
```

packages/ui-components/src/ApplePay/types.ts
```diff
@@ -15,8 +15,13 @@ export interface ApplePayConfig {
   allowedCardNetworks?: string[];
   //These two allow users to add any additional data to the Apple Pay request that we don't currently support
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dave-kav, line 17: You can remove this eslint override now

packages/ui-components/src/ApplePay/utilities.ts
```diff
@@ -59,26 +88,142 @@ export function buildSession(
...
+
   const request = new PaymentRequest(
-    paymentMethodData,
-    paymentDetails,
+    paymentOverrides.paymentMethodData
+      ? paymentOverrides.paymentMethodData
+      : paymentMethodData,
+    paymentOverrides.paymentDetails
+      ? paymentOverrides.paymentDetails
+      : paymentDetails,
```
dave-kav, line 99: It could be possible to leverage the `??` operator here if it makes sense?
```suggestion
    paymentOverrides.paymentMethodData ?? paymentMethodData,
    paymentOverrides.paymentDetails ?? paymentDetails,
```

packages/ui-components/src/GooglePay/index.tsx
```diff
@@ -29,6 +29,11 @@ export function GooglePay({ config }: GooglePayProps) {
   >();
 
   useLayoutEffect(() => {
+    if (config.transaction.type == "disbursement") {
+      console.log("Google Pay does not support disbursment transactions.");
```
dave-kav, line 33: Should we log this as a warn or error?

dave-kav (commented):
Looks good, few small nits/qs

packages/types/uiComponents.ts
donaltuohy: `TransactionDetails` is the type of object the user passes to the wallet components. I created the new type so the user wouldn't be exposed to the new `domain` field, as we add that in the browser SDK. So `TransactionDetails` is what the user specifies and then before we send it to UI Components, we convert it to `TransactionDetailsWithDomain` with the domain set. We could make domain optional in the base transaction either which would allow the user to override it but I don't think there would be a case we would need that. What do you think?

packages/ui-components/src/ApplePay/utilities.ts
donaltuohy: Yeah good shout. I went with `||` however to catch all all falsey values

packages/types/uiComponents.ts
dave-kav: Ah no fair enough, you're right, leave it as is, we don't want to expose that 👍

dave-kav (approved):

PR #534, opened 2025-01-16
Ensure all versioning of workspace deps is consitent

ghost:
# Why

Noticed some inconsistent things while investigating something

# How


donaltuohy (approved):

thomascullen:
Think you need to run `pnpm i` to update the lockfile.

PR #602, opened 2025-03-24
Add support for requesting payer name, email and phone with Apple Pay

thomascullen:
Also includes a new dedicated apple-pay example project for easier dev environment.


packages/ui-components/src/ApplePay/utilities.ts
```diff
@@ -80,10 +80,10 @@ function buildPaymentSession(
...
 
   const paymentOptions = {
-    requestPayerName: false,
+    requestPayerName: config.requestPayerName ?? false,
     requestBillingAddress: false,
-    requestPayerEmail: false,
-    requestPayerPhone: false,
+    requestPayerEmail: config.requestPayerEmail ?? false,
+    requestPayerPhone: config.requestPayerPhone ?? false,
     requestShipping: false,
```
dave-kav, line 87: Is there a case to also expose the billingAdress/shipping info while we're at it? And if so, I wonder if something like below is more/less ergonomic than 5 boolean fields:
```ts
requiredPayerDetails: ["name", "email", "phone"]
...

requestPayerName: config.requiredPayerDetails.includes("name")

```

boilsquid (approved):

dave-kav (commented):
1 quick question

packages/ui-components/src/ApplePay/utilities.ts
thomascullen: Good idea, I had just kept them the same to essentially map the apple pay options.

Should have mentioned re: billing/shipping address. I have left them out as its a little more complicated for them as it requires implementing various methods like [`onshippingaddresschange`](https://www.w3.org/TR/payment-request/#onshippingaddresschange-attribute.)

I will take a look at supporting this with the other work for supporting third party browsers.

dave-kav: Gotcha, lets keep this quick and dirty then since well be re-working a lot of it anyway

PR #609, opened 2025-03-26
Support Apple Pay in third party browsers

thomascullen:
# Why
 Apple Pay now supports third party browsers by rendering a modal window with a scannable code for the user to handle auth on their iPhone. However, this does not work for our current implementation due to it being inside of an iframe. The modal window opens inside of the iframe document and Apple provides no way to control where the modal is mounted.

# How
Rewrite the Apple Pay component to render inside of the parent document. This is only possible because Apple Pay deals with tokens instead of raw PANs and so the compliance requirements do not require an iframe boundary. I've resisted the urge to do much refactoring to try to keep the API contracts and functionality the same as the other implementation.

For now I have nested this under a new `ui.applePayButton` component in order to keep the old implementation around. Once we are happy with the changes we can move it to `ui.applePay` and alias the `ui.applePayButton` method.

Also includes:
- Support for requesting shipping information with the `requestShipping` option.
- Support for requesting billing address with the `requestBillingAddress` option.

The API contract is the exact same as the previous implementation except for one breaking change: Previously we reformatted the `bililngContact` that apple provided us to nest the address fields. Now we just pass the fields through as provided by Apple. This will only affect those using disbursements as before there wasn't a way to request billing details outside of disbursements.

```
Before
{
  ...,
  billingContact: {
    givenName: '...',
    familyName: '...',
    emailAddress: '...',
    phoneNumber: '...'.
    address: { 
      addressLines: [],
      administrativeArea: '',
      country: '',
      ...
  }
}

After: The billing contact as apple provides it
{
  ...,
  billingContact: {
    givenName: '...',
    familyName: '...',
    emailAddress: '...',
    phoneNumber: '...'.
    addressLines: [],
    administrativeArea: '',
    country: '',
    ...
}
```

### DIFF
--- a/packages/browser/lib/ui/ApplePay/index.ts
+++ b/packages/browser/lib/ui/ApplePay/index.ts
@@ -4,94 +4,21 @@ import { resolveSelector } from "../utils";
 import { Transaction } from "../../resources/transaction";
 import { buildSession } from "./utilities";
 import EventManager from "../eventManager";
+import {
+  ApplePayButtonLocale,
+  ApplePayButtonStyle,
+  ApplePayButtonType,
+  ApplePayCardNetwork,
+} from "./types";
+import { tryCatch } from "../../utilities";
 
 const API = import.meta.env.VITE_API_URL || "https://api.evervault.com";
 const APPLE_PAY_SCRIPT_URL =
   "https://applepay.cdn-apple.com/jsapi/1.latest/apple-pay-sdk.js";
 
-type ApplePayButtonType =
-  | "add-money"
-  | "book"
-  | "buy"
-  | "check-out"
-  | "continue"
-  | "contribute"
-  | "donate"
-  | "order"
-  | "pay"
-  | "plain"
-  | "reload"
-  | "rent"
-  | "set-up"
-  | "subscribe"
-  | "support"
-  | "tip"
-  | "top-up";
-
-type ApplePayButtonStyle = "black" | "white" | "white-outline";
-
-type ApplePayButtonLocale =
-  | "ar-AB"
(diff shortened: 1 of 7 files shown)


dave-kav (commented):
Sick! 🚀 🥇 💥 🚢 

image

packages/browser/lib/ui/ApplePay/index.ts
```diff
@@ -0,0 +1,210 @@
...
+  type?: ApplePayButtonType;
+  style?: ApplePayButtonStyle;
+  locale?: ApplePayButtonLocale;
+  padding?: string;
+  borderRadius?: string;
+  size?: { width: string; height: string };
+  allowedCardNetworks?: ApplePayCardNetwork[];
+  requestPayerDetails?: ("name" | "email" | "phone")[];
+  requestBillingAddress?: boolean;
+  requestShipping?: boolean;
```
dave-kav, line 33: Only minor comment is whether you think its worth reconciling these somehow or another?

thomascullen: Yeah I'm still a little on the face about merging these because they are fields actually defined in the [payment request spec](https://developer.mozilla.org/en-US/docs/Web/API/PaymentRequest/PaymentRequest#options).

We could merged them down to `requestDetails: ['name', 'email', 'phone', 'address', 'shipping']` which I think is a nicer API but just feels weird to separate from the spec.

dave-kav: Fair comment. Happy to leave as is if you feel best.

dave-kav (approved):

PR #623, opened 2025-03-28
Isolate build, lint and test into smaller workflows to run in parallel

lfarrel6:
# Why

Our CI for this repo is in a dire state, and needs some love.

# How

First PR of several — rework the PR action:
- Split out linting steps from build
- Run builds in parallel for all packages in this repo
- Make artifact upload conditional
- Make test coverage conditional
- Cache playwright binaries, and run e2e tests in parallel

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,72 @@
+name: Build Package
+on:
+  workflow_call:
+    inputs:
+      vite-evervault-js-url:
+        description: "URL where the JS SDK should be loaded from"
+        required: true
+        type: string
+      vite-keys-url:
+        description: "URL where keys will be fetched from"
+        required: true
+        type: string
+      vite-api-url:
+        description: "API endpoint"
+        required: true
+        type: string
+      environment:
+        description: "The environment that the URLs reference"
+        required: true
+        type: string
+      package-name:
+        description: "The name of the package to build. This must equal the name in the package.json"
+        required: true
+        type: string
+      upload-artifact:
+        description: "Whether to upload the artefact or not"
+        required: true
+        type: boolean
+      artifact-name:
+        description: "The name of the artifact to upload. Will be uploaded with the environment name as a prefix e.g. staging-<artifact-name>"
+        required: false
+        type: string
+jobs:
+  build:
+    name: "[${{ inputs.environment }}] Build ${{ inputs.package-name }}"
+    runs-on: ubuntu-latest
+    env:
+      VITE_API_URL: ${{ inputs.vite-api-url }}
+      VITE_EVERVAULT_JS_URL: ${{ inputs.vite-evervault-js-url}}
+      VITE_KEYS_URL: ${{ inputs.vite-keys-url }}
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Cache turbo build setup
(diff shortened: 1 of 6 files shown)


boilsquid (approved):

.github/workflows/lint.yml
```diff
@@ -0,0 +1,73 @@
...
+      # TODO: resolve issues with unit test coverage
+      # - name: Build in coverage tooling
+      #   if: ${{ inputs.run-code-coverage-checks }}
+      #   run: pnpm run -r --if-present build
+      - name: Formats check
+        run: pnpm run format:check
+      - name: Typescript check
+        run: pnpm run typecheck
+      - name: eslint check
+        run: pnpm run lint
```
thomascullen, line 64: We could probably buy a little more time here by only linting/type checking the files that have changed. Not a big deal but I know the type checking is getting close to a minute now and is likely type checking packages we don't need to be. i.e tests/examples.

Just an FYI these scripts use turbo in the root package.json so could be a lot more going on due to turbo dependencies.

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,166 @@
...
+        with:
+          node-version-file: ".nvmrc"
+          cache: "pnpm"
+      - name: Install deps
+        run: pnpm install --frozen-lockfile
+      - name: Create list of E2E test projects
+        id: e2e-targets
+        run: |-
+          E2E_TEST_PROJECTS=$(pnpm m ls --depth=1 --json | jq -cr '[.[] | select(.name|endswith("e2e-tests")) | .name]')
+          echo "e2e-test-projects=$E2E_TEST_PROJECTS" >> $GITHUB_OUTPUT
```
thomascullen, line 102: Not something for this PR but I feel like rather than depending on the name here we should just read from the e2e-tests dir and move the preflight tests to their own folder.

.github/workflows/build.yml
```diff
@@ -0,0 +1,72 @@
...
+          restore-keys: |
+            ${{ runner.os }}-turbo-
+      - uses: pnpm/action-setup@v4
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: ".nvmrc"
+          cache: "pnpm"
+      - name: Install deps
+        run: pnpm install --frozen-lockfile
```
thomascullen, line 58: Not necessarily a bad thing but the `card-validator` and both react-native packages have prepare scripts causing us to build those packages evey time we install deps. 

Just something I noticed looking at the run here:https://github.com/evervault/evervault-js/actions/runs/14134570449/job/39603065138

thomascullen (approved):

.github/workflows/build.yml
lfarrel6: Oh nice that's a good spot. Will move the prepare script to prebuild which should be hit on the same paths. Should save some time if we manage to get proper change filtering

lfarrel6: should be sorted in latest commit

.github/workflows/e2e-test.yml
lfarrel6: yep agreed

lfarrel6:
@SocketSecurity ignore npm/vite@5.4.11

PR #626, opened 2025-03-31
Refactor push workflow and tidy up

lfarrel6:
# Why

The workflow for a merge to master needs to be updated to use the new modular actions to give us a performance improvement and better readability.

# How

- Update action on main to test all packages
- Parallelise all package builds
- Standardize the artifact naming using an action to translate package names and environments into consistent names
- Delete old actions
- ~Move all invoked actions to live under `.github/workflows/actions` to leave the top level for the automated CI.~ This is not supported by Github Actions..


thomascullen (approved):
All looks good to me! Only thing to point out is that with this I don't think we will be running the e2e tests on push or release. I don't think we actually need them but just calling it out to make sure its intended.

PR #627, opened 2025-03-31
Isolate environments in the js repo

lfarrel6:
# Why

We have gated releases for the browser SDK and ui components, but the environment label is the same for both. This means that they cannot be approved independently. Added package specific prefix to their environment checks.

# How

- Added package specific prefix to browser and ui components environment checks.
- Move pre-release projects from under the e2e tests directly, and update the e2e test workflow to just parse out all projects under `e2e-tests`


lfarrel6:
both ui-components-production-deploy and browser-production-deploy environments have been created and configured to require explicit review

thomascullen (approved):
✨

PR #629, opened 2025-04-01
Consolidate invoked workflows on push

lfarrel6:
# Why

Exceeding Github's threshold of invoked workflows in `push.yml` so need to consolidate them.

# How

- Implement a `publish-static-bundle.yml` workflow which standardizes the S3 upload and cloudfront invalidation process
- Remove all package specific release workflows
- Migrate push to use new consolidated workflow


thomascullen (approved):
🤞

PR #642, opened 2025-04-14
Improve Apple Pay availability checks

thomascullen:
# Why

Apple Pay must meet several requirements for the user to be able to use it, e.g., set up payment methods, their laptop must be open so biometric keys are available, etc. Currently, we don't provide a nice way to check for the availability of Apple Pay. We also don't handle certain scenarios, such as biometric keys not being available, which leads to a button being rendered that the user can not use.

# How

Adds a new `availability` method to determine whether or not Apple Pay is available. This method will respond with one of three possible strings.

- "available": Apple Pay is available and can be used.
- "unavailable": Apple Pay is not available due to payment credentials being unavailable.
- "unsupported": Apple Pay is not supported on this device or browser.

Update the `mount` method to check the result of the `availability` method and prevent rendering the button if Apple Pay availability is 'unsupported'


packages/browser/lib/ui/ApplePay/index.ts
```diff
@@ -166,7 +171,69 @@ export default class ApplePayButton {
...
+
+    if (availability === "UNSUPPORTED") {
+      console.info("Apple pay is not supported on this device.");
+      return;
+    }
+
+    if (availability === "UNAVAILABLE") {
+      console.info("Apple pay is currently uavailable on this device.");
+      return;
+    }
```
thomascullen, line 235: Would it make more sense to throw an error here or is it better to fail more silently and encourage the use of the `availability` method to programmatically handle availability?

dave-kav: Can we fire an bespoke event? Force the logic onto the consumer, e.g. on an "unavailable" event user's can choose to do nothing or display their own message/error/warning whatever?

packages/browser/lib/ui/ApplePay/index.ts
```diff
@@ -166,7 +171,69 @@ export default class ApplePayButton {
...
+  async mount(selector: SelectorType) {
+    const availability = await this.availability();
+
+    if (availability === "UNSUPPORTED") {
+      console.info("Apple pay is not supported on this device.");
+      return;
+    }
+
+    if (availability === "UNAVAILABLE") {
+      console.info("Apple pay is currently uavailable on this device.");
```
dave-kav, line 233:
```suggestion
      console.info("Apple pay is currently uanvailable on this device.");
```

PR #645, opened 2025-04-17
fix: get canonical artifact name in publish flow

lfarrel6:
# Why

Release of the react native SDK failed because the artifact path was not updated in our CI overhaul. It should be using the composable workflow to get a canonical artifact name.

# How

Add call to `get-artifact-name` in npm publish workflow. It might be useful to update this workflow to perform a dry run publish and have the artifact available for inspection in future.


joshpensky (approved):
let's try it! 🚀

PR #647, opened 2025-04-18
fix: parse out the changelog relevant to the specific release

lfarrel6:
# Why

Releases being created on publish contain the entirety of a packages changes instead of the details for the release in question

# How

For a release commit, take the diff of the changelog file, strip the diff formatting, and use it as the changelog content.


thomascullen (approved):
noice

joshpensky (approved):
so good 👏

PR #660, opened 2025-04-24
fix: correct prerelease steps in workflows to run the correct harness

lfarrel6:
# Why

Pre-release tests are running against the wrong projects

# How

- Update workflow to split between browser and ui-components
- Refine changelog logic

### DIFF
--- a/.github/workflows/publish-static-bundle.yml
+++ b/.github/workflows/publish-static-bundle.yml
@@ -142,18 +142,34 @@ jobs:
         run: pnpm install --frozen-lockfile
       - name: Install Playwright Browsers
         run: pnpm exec playwright install --with-deps
-      - name: Run preflight checks
-        if: ${{ inputs.run-preflight-checks }}
-        id: preflight
+      - name: Run browser preflight checks
+        if: ${{ inputs.run-preflight-checks && inputs.package-name == '@evervault/browser' }}
+        id: browser-preflight
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-static-bundle.yml
```diff
@@ -142,18 +142,34 @@ jobs:
...
+          service_name: ${{ inputs.package-name }} Preflight check failed for ${{ inputs.stage }}
+          action: failed
+      - name: Run ui-components preflight checks
+        if: ${{ inputs.run-preflight-checks && inputs.package-name == '@evervault/ui-components' }}
+        id: ui-components-preflight
+        env:
+          VITE_EV_TEAM_UUID: ${{ secrets.TESTS_TEAM_UUID }}
+          VITE_EV_APP_UUID: ${{ secrets.TESTS_APP_UUID }}
+          VITE_API_KEY: ${{ secrets.TESTS_DECRYPT_FN_KEY }}
+          VITE_EVERVAULT_JS_URL: https://ui-components.evervault.${{ inputs.stage == 'production' && 'com' || 'io' }}${{ inputs.cloudfront-preview-path-invalidation }}
```
thomascullen, line 168: I think this should either be set to https://js.evervault.com/v2 or not set at all for ui-comp pre release tests

lfarrel6: good catch! removed JS URL, and set ui comp

PR #671, opened 2025-04-28
Improve error handling for UI components in the react sdk

thomascullen:
# Why

Currently we don't provide a clean way for users to hook into cases when our JS SDK fails to load after being injected into the DOM.

# How

- Add a new `onLoadError` prop to the `EvervaultProvider` component which will be called if the injected script fails to load.
- Call the `onError` callback prop that is passed to the `Card` component if the Evervault SDK is not available when it tried to mount.


packages/react/lib/ui/Card.tsx
```diff
@@ -153,21 +158,26 @@ export function Card({
...
+      try {
+        if (initialized.current || !ref.current) return;
+        initialized.current = true;
+        const evervault = await ev;
+        if (!evervault) return;
+        const inst = evervault.ui.card(config);
+        inst.mount(ref.current);
+        setInstance(inst);
+      } catch (err) {
+        onErrorRef.current?.();
```
joshpensky, line 170: do we want to pass the error to `onError` and `onLoadError`? or is it not really helpful to have it

joshpensky (approved):
Looks good! 🚀 just one non-blocking q

packages/react/lib/ui/Card.tsx
thomascullen: Yeah I was considering this but not sure it would be very useful just yet. I think if we do pass an argument as well we should construct our own error with enums instead as it could be an issue from inside the iFrame leading to the load error.

PR #679, opened 2025-05-07
Add billing address support to Google Pay

thomascullen:
Adds support for requesting billing address information with Google Pay.

You can now collect billing address information using the `billingAddress` option.

```js
const googlePay = evervault.ui.googlePay(transaction, {
    billingAddress: true,
    process: async (payload) => {
        console.log(payload.billingAddress)
    }
});
```

You can also specific the address format and request a phone number by using an object instead of a boolean.

```js
const googlePay = evervault.ui.googlePay(transaction, {
    billingAddress: {
        format: 'MIN',
        phoneNumber: true
    },
    process: async (payload) => {
        console.log(payload.billingAddress)
    }
});
```

### DIFF
--- a/.changeset/strong-kangaroos-reflect.md
+++ b/.changeset/strong-kangaroos-reflect.md
@@ -1,0 +1,32 @@
+---
+"@evervault/ui-components": minor
+"@evervault/browser": minor
+"types": minor
+---
+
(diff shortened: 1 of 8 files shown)


joshpensky (approved):
nice! 🏠

PR #685, opened 2025-05-12
PAY-12 Apple Pay Recurring Payments

dave-kav:
# Why

Adding support for recurring transactions in Apple Pay

Main modal:
image

Information tab:
image


packages/ui-components/src/ApplePay/utilities.ts
```diff
@@ -100,6 +103,93 @@ function buildPaymentSession(
   return request;
 }
 
+function buildRecurringSession(
+  merchant: MerchantDetail,
+  config: ApplePayConfig,
+  tx: RecurringTransactionDetails
+) {
+  console.log("Building recurring session");
```
thomascullen, line 111: Another left over here, although this is the old ui-component which we have deprecated and need to remove this code anyway so I would just remove the changes from ui-components and update the change set.

thomascullen (changes requested):
Just some left over console logs but looks good other than that!

packages/ui-components/src/ApplePay/utilities.ts
dave-kav: Hmm - I'll need to at least add a type assertion in this file, otherwise I end up with a
`Type '"recurring"' is not assignable to type '"disbursement"'.` error

dave-kav: Do we clean up the old component first, or just generate the changeset for only the relevant packages excluding this and any knock-ons

dave-kav: @thomascullen I reverted the old ui-component, generated a changeset for the relevant changes, then added the minimum changes required to the old ui-component in order to satisfy tscheck - is that reasonable? Would be nice if we can get rid of the old component, happy to do so also

dave-kav: 🗑️

PR #690, opened 2025-05-19
PAY-94 Expose last 4 of underlying card for Google Pay

dave-kav:
# Why

We have access to it, at least one customer wants it, so lets expose it.

# How

Set the card.lastFour field of the Google Pay object from `data.paymentMethodData.cardDetails`. This should be the lastFour. If anything other than 4 digits, we try to parse 4 digits from this string. If that fails fallback to attempting to parse if from `paymentMethodData?.description` which also usually contains the last four.

### DIFF
--- a/examples/react-google-wallet/src/App.tsx
+++ b/examples/react-google-wallet/src/App.tsx
@@ -20,7 +20,7 @@ function App() {
         amount: 4300,
         currency: "USD",
         country: "US",
-        merchantId: "merchant_e930d3f7bf37",
+        merchantId: import.meta.env.VITE_MERCHANT_ID,
       });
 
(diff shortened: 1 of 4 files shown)


examples/react-google-wallet/src/App.tsx
```diff
@@ -19,7 +19,7 @@ function App() {
         amount: 4300,
         currency: "USD",
         country: "US",
-        merchantId: "merchant_e930d3f7bf37",
+        merchantId: import.meta.env.VITE_MERCHANT_ID,
```
jakekgrog, line 22: 🍾

jakekgrog (approved):

thomascullen (approved):
💜

PR #691, opened 2025-05-21
Allow for browser SDK load retries in `@evervault/react`

joshpensky:
# Why

Previously, our browser SDK loading logic would never retry if it failed to load once. This is an issue if the script fails to load initially but is then made available later

~While this PR does not add explicit functionality to force a retry, it does allow for retries when the `EvervaultProvider` component is remounted (e.g. a `key` is changed)~

**Edit:** in addition to automatic retries on mount, we've also added a `reload` function to the `EvervaultProvider` ref so the script can retry a load without needing to remount the tree

# How

- Move browser script loading to separate file, add handy hook `useEveraultClient`
- Move to a promise map to allow for multiple load promises to happen concurrently
- Adds new `reload` function via an `EvervaultProvider` ref for manual reloads
- Add new `useEvInstance` hook to consolidate instance initialization in components
- Update `Card`, `Pin`, `Reveal`, and `ThreeDSecure` to use new `useEvInstance` hook
- Update our React example with retry mechanism

### DIFF
--- a/.changeset/mean-masks-burn.md
+++ b/.changeset/mean-masks-burn.md
@@ -1,0 +1,5 @@
+---
+"@evervault/react": minor
+---
+
+Allow retry when Evervault browser SDK fails to load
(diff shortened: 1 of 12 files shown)


thomascullen (approved):
Very nice! 👏

PR #697, opened 2025-05-27
PAY-28 Add prepareTransaction callback for Apple Pay

dave-kav:
# Why

We should allow a user to implement logic to update the transaction before showing the Apple Pay modal, for example to offer a sign-up discount or other promotion.

# How

Add an `onClick` callback to the Apple Pay config, which returns a Promise of `{ amount?: number; lineItems?: TransactionLineItem[]; }` which are used to update the transaction.

See video below for an example implementation:

https://github.com/user-attachments/assets/0e208e57-6e2b-44e1-b566-c66fec510fa4

### DIFF
--- a/packages/browser/lib/ui/ApplePay/index.ts
+++ b/packages/browser/lib/ui/ApplePay/index.ts
@@ -41,6 +41,10 @@ export type ApplePayButtonOptions = {
   onShippingAddressChange?: (
     event: PaymentRequestUpdateEvent
   ) => Promise<{ amount: number; lineItems?: TransactionLineItem[] }>;
+  prepareTransaction?: () => Promise<{
+    amount?: number;
+    lineItems?: TransactionLineItem[];
+  }>;
   process: (
     data: EncryptedApplePayData,
     helpers: {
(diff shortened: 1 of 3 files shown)


packages/browser/lib/ui/ApplePay/index.ts
```diff
@@ -103,8 +107,25 @@ export default class ApplePayButton {
...
+      }
+      const existingLabels = new Set(
+        (this.transaction.details.lineItems ?? []).map((item) => item.label)
+      );
+
+      const lineItemsToAdd = (lineItems ?? []).filter(
+        (item) => !existingLabels.has(item.label)
+      );
+
+      this.transaction.details.lineItems?.push(...lineItemsToAdd);
```
thomascullen, line 126: I think it would be fair to require users to provide the full line items list again rather than checking which ones are new. This would also allow users to remove line items if they wanted to. Could imagine something like a line item being removed/changed based on a coupon code or something

thomascullen (commented):
One small suggestion.

PR #699, opened 2025-05-27
Prevent double creq submission

joshpensky:
# Why

In order to prevent double creq submission, this PR:
- moves the frame initialization calls around (and cleans up some React code) in the `BrowserFingerprint` and `ChallengeFrame` components
- Cleans up the effects and callbacks in `ThreeDSecure` component

# How

- Move `postRedirectFrame` inside a ref callback, so it's called immediately upon node mount
- Move frame `load` listener from the useEffect to the iframe `onLoad` prop
- Removes the window event listener in the useEffect
- Fix timeout references in useEffect
- Separate the status effects in `ThreeDSecure`
- Wrap all function props in `useCallback` for referential integrity


dave-kav (approved):
✨

PR #716, opened 2025-07-04
Expose onpaymentmethodchange event for Apple Pay

dave-kav:
# Why

We have at least one customer with a use case for responding to payment method changes

# How

Expose a hook for this event so that the consumer can e.g. update tax based on location


thomascullen (approved):
Nicely done.

PR #717, opened 2025-07-04
Expose BIN on card onchange event as soon as identifiable

dave-kav:
# Why

We should return the BIN as soon the input length is sufficient so that BIN dependent use cases can be enabled without collecting the full card


thomascullen (commented):
Looks good to me, just need to update the tests.

PR #731, opened 2025-07-31
Feature: Test Sandbox Apps For Google Pay

matt-evervault:
# Why
Add a short description about this PR.
Add links to issues, tech plans etc.

# How

### DIFF
--- a/packages/ui-components/src/GooglePay/index.tsx
+++ b/packages/ui-components/src/GooglePay/index.tsx
@@ -8,6 +8,7 @@ import { GooglePayClientMessages, GooglePayHostMessages } from "types";
 import { useSearchParams } from "../utilities/useSearchParams";
 import { getMerchant } from "../utilities/useMerchant";
 import { getAppSDKConfig } from "../utilities/getAppSDKConfig";
+import { apiConfig } from "../utilities/config";
 
 interface GooglePayProps {
   config: GooglePayConfig;
@@ -40,7 +41,13 @@ export function GooglePay({ config }: GooglePayProps) {
     async function onLoad() {
       const appConfig = await getAppSDKConfig(app);
       const paymentsClient = new google.payments.api.PaymentsClient({
-        environment: appConfig.isSandbox ? "TEST" : "PRODUCTION",
+        // Always use 'test' in staging, but use the resolved environment in production
+        environment:
+          apiConfig.environment === "staging"
+            ? "TEST"
+            : appConfig.isSandbox
+            ? "TEST"
+            : "PRODUCTION",
         paymentDataCallbacks: {
           onPaymentAuthorized: async (data) => {
             const payload = await exchangePaymentData(
--- a/packages/ui-components/src/GooglePay/types.ts
+++ b/packages/ui-components/src/GooglePay/types.ts
@@ -14,5 +14,4 @@ export interface GooglePayConfig {
   allowedAuthMethods?: string[];
   allowedCardNetworks?: string[];
   billingAddress?: GooglePayBillingAddressConfig;
-  environment?: "PRODUCTION" | "TEST";
 }


packages/browser/lib/main.ts
```diff
@@ -80,7 +80,8 @@ export default class EvervaultClient {
       teamId,
       appId,
       customConfig?.urls,
-      customConfig?.publicKey
+      customConfig?.publicKey,
+      customConfig?.appKey?.isSandbox
```
matt-evervault, line 84: Don't love that the isSandbox comes from the appKey, but it's a bit inconvenient to do it otherwise, thinking a comment here could help.

dave-kav: I don't think we can rely on this because CustomConfig is unlikely to be defined in most cases I think. Let me take a look at what might work

matt-evervault:
Tested locally and all seems to be good on both calling the oxide endpoint and rendering google pay in the sample app.

packages/ui-components/src/utilities/getAppSDKConfig.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (!response.ok) {
+    console.error(
+      `Failed to fetch app SDK config details for ${app}`,
+      response.status
+    );
+    return getDefaultAppSDKConfig();
+  }
+
+  return response.json() as Promise<AppSDKConfig>;
+}
```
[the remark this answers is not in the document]
matt-evervault: ✅

packages/ui-components/src/GooglePay/types.ts
```diff
@@ -14,4 +14,5 @@ export interface GooglePayConfig {
   allowedAuthMethods?: string[];
   allowedCardNetworks?: string[];
   billingAddress?: GooglePayBillingAddressConfig;
+  environment?: "PRODUCTION" | "TEST";
```
dave-kav, line 17: I think we can omit this `environment` type definition because I don't think we use this elsewhere now, we just set the environment right in the `google.payments.api.PaymentsClient` constructor

packages/ui-components/src/GooglePay/index.tsx
```diff
@@ -38,8 +38,9 @@ export function GooglePay({ config }: GooglePayProps) {
     called.current = true;
 
     async function onLoad() {
+      const appConfig = await getAppSDKConfig(app);
       const paymentsClient = new google.payments.api.PaymentsClient({
-        environment: apiConfig.googlePayEnvironment,
+        environment: appConfig.isSandbox ? "TEST" : "PRODUCTION",
```
dave-kav, line 43: I think we potentially still need an additional check for staging vs production - for staging we should maintain the current behaviour where we always use the google "TEST" environment. For production, we apply this logic.

dave-kav (commented):
Couple small things

packages/ui-components/src/GooglePay/index.tsx
```diff
@@ -40,7 +41,13 @@ export function GooglePay({ config }: GooglePayProps) {
...
       const appConfig = await getAppSDKConfig(app);
       const paymentsClient = new google.payments.api.PaymentsClient({
-        environment: appConfig.isSandbox ? "TEST" : "PRODUCTION",
+        // Always use 'test' in staging, but use the resolved environment in production
+        environment:
+          apiConfig.environment === "staging"
+            ? "TEST"
+            : appConfig.isSandbox
+            ? "TEST"
+            : "PRODUCTION",
```
dave-kav, line 50:
```suggestion
environment: (appConfig.isSandbox || !apiConfig.environment === 'production') ? "TEST" : "PRODUCTION"
```

I'd be inclined to avoid the nested ternary, but that's purely preference, non-blocking

dave-kav (approved):

packages/ui-components/src/GooglePay/index.tsx
matt-evervault: I think while it's less code, I'd argue it hides the logic we want to represent.  I think I'll leave it as-is for now.

PR #747, opened 2025-08-21
Initial commit remove rollup

matt-evervault:
# Why
As part of react native work we are doing for evervault-pay.  We think it would be simpler to do that if we didn't have rollup.  We want to test that there, then we can rebase that work on top of this. #725 

# How


matt-evervault:
Superseded by: https://github.com/evervault/evervault-js/pull/754

PR #748, opened 2025-08-22
Vibe Coded Auto Format

matt-evervault:
# Why
Formatting vibes

# How


joshpensky (approved):
This feels right to me! 🤘

@thomascullen Matt and I were discussing removing the linting from commits and adding it to CI instead

thomascullen:
> This feels right to me! 🤘
> 
> @thomascullen Matt and I were discussing removing the linting from commits and adding it to CI instead

@joshpensky @matt-evervault I think that makes sense for things that can be resolved in CI like formatting, however, I think it makes sense to keep things like typechecks on a pre commit hook. They can also be bypassed with ` git commit --no-verify` if need be.

PR #754, opened 2025-08-28
Support platform-specific extensions in RN bundle

joshpensky:
# Why

Screenshot 2025-08-28 at 09 16 49

We need to support platform-specific extensions in `@evervault/react-native` in order to handle separate ApplePay and GooglePay components

However, we can't simply transpile TS to JS since we also have third-party dependencies we don't want to expose as peers. So, the solution is we need to bundle our files while keeping the same file structure so Metro can properly resolve platform-specific files

# How

- Update Rollup config to:
    - use a glob for the `import` (so it catches `.ios.tsx`/`android.tsx` files even though they're not explicitly referenced in code)
    - preserve module structure (keeping all files in the place they're declared)
    - include a new `platformResolution` plugin to exclude explicit `index.js` imports (so Metro can do its magic)
- Update package.json exports
- Add a temporary `ApplePayButton` for testing platform-specific extensions (this should be removed before PR merge)

# Gotchas

- Rollup doesn't support native glob imports, so we need to use the `glob` package. Unfortunately, this doesn't play well with watch mode since the glob is only evaluated when the command is first run. That means new files aren't detected until the command is rerun, and deleted files will break the build because they can no longer be found


pnpm-workspace.yaml
joshpensky, line 1: My Xcode wasn't playing nice with the old versions we were on. These are minor updates with some patches so Xcode doesn't freak out!

PR #809, opened 2026-02-02
Skip CVC validation when card number validation fails

joshpensky:
# Why

Fixes a bug where CVC validation fails automatically when card number validation fails

Since CVC validation is based on card number, we should wait until the card number is valid first

### Before

https://github.com/user-attachments/assets/bb416620-068e-49d2-8caf-6b80194aeee2

### After

https://github.com/user-attachments/assets/49d7edc8-650c-4514-982e-647221fd9a6a

# How

* Update CVC validation to skip validation if `cardValidation.isValid` is false

### DIFF
--- a/.changeset/spicy-mayflies-give.md
+++ b/.changeset/spicy-mayflies-give.md
@@ -1,0 +1,7 @@
+---
+"@evervault/card-validator": minor
+"@evervault/ui-components": patch
+---
+
+- Adds `reason` to CVC validation failures
+- Fixes bug where failed card validation forces a failed CVC validation
--- a/.changeset/spicy-mayflies-give.md
+++ b/.changeset/spicy-mayflies-give.md
@@ -1,7 +1,6 @@
 ---
-"@evervault/card-validator": minor
+"@evervault/card-validator": patch
 "@evervault/ui-components": patch
 ---
 
-- Adds `reason` to CVC validation failures
-- Fixes bug where failed card validation forces a failed CVC validation
+- Fixes bug where failed card number validation forces a failed CVC validation
--- a/packages/card-validator/index.ts
+++ b/packages/card-validator/index.ts
@@ -128,8 +128,6 @@ export function validateCVC(
   cardNumber?: string,
   options?: CardNumberValidationOptions
 ): CardCVCValidationResult {
-  const { brand, localBrands } = validateNumber(cardNumber ?? "", options);
-
   // Check if the CVC only contains numbers with 3 or 4 digits
   if (!/^\d{3,4}$/.test(cvc)) {
     return {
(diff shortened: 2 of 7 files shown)


matt-evervault (approved):

packages/ui-components/src/Card/index.tsx
```diff
@@ -106,8 +106,15 @@ export function Card({ config }: { config: CardConfig }) {
...
         const cvcValidation = validateCVC(values.cvc, values.number);
-
         if (!cvcValidation.isValid) {
+          // Skip CVC validation if failed because of invalid card number
+          if (
+            cvcValidation.reason === "invalid_number" &&
+            !cardValidation.isValid
+          ) {
+            return undefined;
+          }
```
jmorgadosoares: I might be splitting hairs, and maybe overcomplicating so feel free to ignore

But I reckon when a CVC is entered that would be invalid for any card type (basically any CVC <3 digits), we should persist the invalid warning. However, if the CVC could be valid depending on the card type, that's when we should clear the warning.

On another note, would it be better to try to decouple CVC validation in the `card-validator` from card number validation? It feels here we are hijacking the `validateCVC` behaviour, instead of improving it

packages/card-validator/index.ts
```diff
@@ -129,6 +130,7 @@ export function validateCVC(
     return {
       cvc: null,
       isValid: false,
+      reason: "invalid_number",
```
jmorgadosoares: same as my comment bellow. I wonder if CVC validation should just be decoupled from card validation.

The reason being is that conceptually the `validateCVC` function should return `true` or `false` only. Imo the card number being valid should be irrelevant (the brand the card number matches is a different story)

pseudo-code cvc validation suggestion:
```
const { brands } = validateNumber(cardNumber); // used only to check if number matches any brands

if (cvc valid for any brands) {
  return true;
}

if (brands && cvc not valid for any brands) {
  return false;
}

if (no brands && cvc.length < 3) {
  return false;
}

return true;
```

packages/shared/src/useForm.ts
```diff
@@ -166,7 +166,12 @@ export function useForm<T extends object>({
       }
     ) => {
       const handleBlur = (e: React.FocusEvent<HTMLInputElement>) => {
-        validateField(name);
+        const value = values[name];
+        if (typeof value === "string" && value.length === 0) {
+          setError(name, undefined);
+        } else {
+          validateField(name);
+        }
```
joshpensky: I would address this bit separately since this changes our current validation functionality quite a bit

This is something we can explore via a `validationMode` prop that users can customize like in our React Native SDK

packages/card-validator/index.ts
```diff
@@ -128,6 +128,8 @@ export function validateCVC(
   cardNumber?: string,
   options?: CardNumberValidationOptions
 ): CardCVCValidationResult {
+  const { brand, localBrands } = validateNumber(cardNumber ?? "", options);
```
joshpensky: **nit:** I'd move this after the regex check since it's not needed till then

packages/card-validator/index.ts
```diff
@@ -136,39 +138,41 @@ export function validateCVC(
...
+    return {
+      cvc,
+      isValid: true,
+    };
   }
 
   const allBrands = [...defaultBrands, ...(options?.customBrands ?? [])];
+
   const isCVCValid = allBrands
     .filter((brand) => brands.includes(brand.name))
```
joshpensky: **nit:** we could make `brands` a Set to reduce some loop cycles here?

.changeset/spicy-mayflies-give.md
joshpensky, line 1: This changeset should be updated from the changes

joshpensky (commented):
thanks for taking this! left a few comments about splitting this out

PR #871, opened 2026-05-19
Append sandbox label to Apple Pay merchant name

maxharrison:
# How
For Apple Pay, when the app is flagged as sandbox, append "(Card is not charged)" to the merchant name displayed in the Apple Pay sheet.

# Why
So test transactions are visibly distinguished from live ones.

# Other changes in this PR
- Moved `getAppSDKConfig` from `ui-components` into `shared` so Apple Pay and Google Pay share it. See comment on `packages/shared/package.json` re: the `exports` field.
- Fixed a latent bug in `ui-components` GooglePay: API URL now comes from `apiConfig.apiUrl` (production-aware) instead of `VITE_API_URL` (could be undefined in production builds).
- Added vitest coverage: 3 cases for `getAppSDKConfig`, 2 for the Apple Pay label append behaviour.

### DIFF
--- a/packages/browser/lib/ui/ApplePay/utilities.ts
+++ b/packages/browser/lib/ui/ApplePay/utilities.ts
@@ -1,4 +1,5 @@
 import {
+  AppSDKConfig,
   DisbursementTransactionDetails,
   MerchantDetail,
   PaymentTransactionDetails,
@@ -52,6 +53,14 @@ export async function buildSession(
   const merchant = await getMerchant(applePay, tx.merchantId);
   if (!merchant) {
     throw new Error("Merchant not found");
+  }
+
+  const appConfig = await getAppSDKConfig(
+    applePay.client.config.appId,
+    applePay.client.config.http.apiUrl
+  );
+  if (appConfig.is_sandbox) {
+    merchant.name = `${merchant.name} (Card is not charged)`;
   }
 
   let baseRequest: ApplePayPaymentRequest;
@@ -509,6 +518,30 @@ async function getMerchant(
   }
 
   return response.json();
+}
+
+async function getAppSDKConfig(
+  app: string,
+  apiUrl: string
+): Promise<AppSDKConfig> {
+  try {
+    const response = await fetch(`${apiUrl}/frontend/sdk/config`, {
+      method: "GET",
+      headers: {
+        "Content-Type": "application/json",
+        "X-Evervault-App-Id": app,
+      },
+    });
+
+    if (response.ok) {
+      return response.json() as Promise<AppSDKConfig>;
+    }
+
+    console.error(`Failed to fetch app SDK config details for ${app}`);
+  } catch (error) {
+    console.error(`Failed to fetch app SDK config details for ${app}`, error);
+  }
+  return { is_sandbox: false };
 }
 
 export function resolveUnit(input: string | number) {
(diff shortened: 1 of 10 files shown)


packages/browser/lib/ui/ApplePay/utilities.ts
```diff
@@ -511,6 +520,30 @@ async function getMerchant(
   return response.json();
 }
 
+async function getAppSDKConfig(
```
maxharrison: This is just a copy of `packages/ui-components/src/utilities/getAppSDKConfig.ts:11` - so I could move it to a shared function maybe in `packages/shared`?

packages/browser/lib/ui/ApplePay/utilities.ts
```diff
@@ -54,6 +55,14 @@ export async function buildSession(
     throw new Error("Merchant not found");
   }
 
+  const appConfig = await getAppSDKConfig(
```
maxharrison, line 58: Could do this in parallel with the `getMerchant` request above with a `Promise.all`, but not sure if its needed tbh.

packages/browser/lib/ui/ApplePay/utilities.ts
maxharrison: Yep did it in `95e7097`

packages/shared/package.json
```diff
@@ -3,6 +3,10 @@
   "private": true,
   "version": "1.1.21",
   "main": "src/index.ts",
+  "exports": {
```
maxharrison, line 6: Needed these exports lines because importing `getAppSDKConfig` from shared was accidentally pulling React into the browser SDK bundle (since shared also exports React hooks).

Cleaner alternative: `"sideEffects": false` on `shared/package.json`, would let us drop these lines and use the normal import. Tradeoff is it's a blanket promise that nothing in shared runs at import time, which is easy to silently break later.

Or: just duplicate the helper into Apple Pay and skip shared altogether. Any preference?

packages/shared/src/getAppSDKConfig.ts
```diff
@@ -19,7 +20,7 @@ export async function getAppSDKConfig(app: string): Promise<AppSDKConfig> {
     });
 
     if (response.ok) {
-      return response.json() as Promise<AppSDKConfig>;
+      return (await response.json()) as AppSDKConfig;
```
jakekgrog, line 23: Do we need to `await` here? This function returns a `Promise` that can be awaited on by the caller

packages/browser/lib/ui/ApplePay/utilities.ts
jakekgrog: For two calls I think this is fine as is

jakekgrog (changes requested):

packages/shared/package.json
joshpensky: Adding the `exports` makes sense to me! Would rather pull this into a shared lib than duplicate

packages/shared/src/getAppSDKConfig.ts
maxharrison: Could return the `Promise` directly, but then a `response.json()` promise rejection would be outside the try catch block

jakekgrog (approved):

PR #884, opened 2026-05-29
Prevent 3DS callback events from firing multiple times

jmorgadosoares:
# Why
3DS callback events (`onSuccess`, `onFailure`, `onError`) could fire multiple times for a single session. The root cause was users spamming the cancel button, which sent multiple `EV_CANCEL` messages, each triggering a separate PATCH to update the session outcome and dispatching the user callback again.

Fixes EXP-992.

# How
Three layers of defence:

- **Idempotency guard on `#handleOutcome`** — a `#handled` flag ensures only the first outcome (success, failure, or cancel) is processed, regardless of how many duplicate events arrive
- **Early listener removal in trampoline handlers** — `BrowserFingerprint` and `ChallengeFrame` now remove their `message` listener before acting on the trampoline event, consistent with how the existing timeout path already behaved
- **Cancel button disabled after first click** — the overlay cancel button switches to `Cancelling...` and becomes disabled immediately on click, preventing duplicate `EV_CANCEL` events from being sent

### DIFF
--- a/packages/browser/lib/ui/threeDSecure.ts
+++ b/packages/browser/lib/ui/threeDSecure.ts
@@ -27,6 +27,7 @@ export default class ThreeDSecure {
   >;
 
   #events = new EventManager<ThreeDSecureEvents>();
+  #handled = false;
 
   constructor(
     client: EvervaultClient,
@@ -92,6 +93,8 @@ export default class ThreeDSecure {
     cres?: string | null,
     abortedOnChallenge: boolean = false
   ) {
+    if (this.#handled) return;
+    this.#handled = true;
     await this.#updateOutcome(outcome, cres, abortedOnChallenge);
     this.#events.dispatch(outcome === "success" ? "success" : "failure");
     this.unmount();
--- a/packages/browser/lib/ui/threeDSecure.ts
+++ b/packages/browser/lib/ui/threeDSecure.ts
@@ -95,7 +95,12 @@ export default class ThreeDSecure {
   ) {
     if (this.#handled) return;
     this.#handled = true;
-    await this.#updateOutcome(outcome, cres, abortedOnChallenge);
+    try {
+      await this.#updateOutcome(outcome, cres, abortedOnChallenge);
+    } catch {
+      this.#handled = false;
+      return;
+    }
     this.#events.dispatch(outcome === "success" ? "success" : "failure");
     this.unmount();
   }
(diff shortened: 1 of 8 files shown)


jakekgrog (approved):

packages/ui-components/src/ThreeDSecure/Overlay.tsx
```diff
@@ -30,7 +41,7 @@ export function Overlay({
               d="M18 6L6 18M6 6l12 12"
             />
           </svg>
-          Cancel
+          {cancelling ? "Cancelling..." : "Cancel"}
```
joshpensky: I'd avoid changing this text for now to keep parity

packages/browser/lib/ui/threeDSecure.ts
```diff
@@ -92,6 +93,8 @@ export default class ThreeDSecure {
     cres?: string | null,
     abortedOnChallenge: boolean = false
   ) {
+    if (this.#handled) return;
+    this.#handled = true;
     await this.#updateOutcome(outcome, cres, abortedOnChallenge);
```
joshpensky: bit of defensive coding: since this call is async it could technically fail and then we wouldn't be able to retry the outcome

worth adding a `try/catch` here?

packages/browser/lib/ui/threeDSecure.ts
joshpensky, line 1: we should add some tests here to check that we can't double-submit outcomes

joshpensky (changes requested):
nice work! this'll def fix some edge case-y bugs 🐞

left a few comments and suggestions

packages/browser/test/threeDSecure.test.ts
```diff
@@ -0,0 +1,241 @@
...
+  server.resetHandlers();
+  vi.restoreAllMocks();
+});
+
+afterAll(() => {
+  server.close();
+});
+
+describe("ThreeDSecure#handleOutcome idempotency", () => {
+  it("only calls PATCH once when EV_CANCEL fires multiple times", async () => {
```
joshpensky, line 102: we should add tests for
- if a cancel then success/failure event happens
- if a success/failure then cancel event happens

joshpensky (commented):
very nice! 👨‍🔬 just a couple more tests

packages/browser/test/threeDSecure.test.ts
jmorgadosoares: Added three more tests tests: cancel→success, success→cancel, and failure→cancel.

joshpensky (approved):
awesome! 🌟

PR #885, opened 2026-05-29
Add custom brand support to card-validator

jmorgadosoares:
# Why
Adds support for custom card brand definitions in `@evervault/card-validator`. Part of the Custom Brands RFC.

Linear: https://linear.app/evervault/issue/CARD-724

# How
- Added a `Brand` type with an arbitrary `name: string` (as opposed to the closed `CardBrandName` union) alongside `NumberValidationRules` and `SecurityCodeValidationRules`
- Added `CardNumberValidationOptions` (`{ customBrands?: Brand[] }`) as an optional second argument to `validateNumber` and `validateCVC` — custom brands are merged with the built-in defaults at validation time
- Tightened `NumberValidationRules.ranges` from `number[] | [number[]]` to `Array<number | [number, number]>` now that this type is part of the public API
- Widened `CardNumberValidationResult.brand` and `localBrands` from `CardBrandName` to `string` to accommodate arbitrary custom brand names in results, with corresponding fixes in `react-native` and `ui-components`

### DIFF
--- a/packages/card-validator/types.ts
+++ b/packages/card-validator/types.ts
@@ -2,7 +2,7 @@ import { CardBrandName } from "types";
 
 export type NumberValidationRules = {
   luhnCheck: boolean;
-  ranges: number[] | [number[]];
+  ranges: Array<number | [number, number]>;
   lengths: number[];
 };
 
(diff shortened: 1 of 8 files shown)


packages/card-validator/types.ts
```diff
@@ -2,28 +2,34 @@ import { CardBrandName } from "types";
...
-  ranges: number[] | [number[]];
+  ranges: Array<number | [number, number]>;
   lengths: number[];
 };
 
 export type SecurityCodeValidationRules = {
   lengths: number[];
 };
 
-export type CardBrand = {
```
joshpensky, line 13: since these types are publicly exported from the package, I would be wary of removing any in a minor version

maybe we just re-export `DefaultBrand` as `CardBrand`?

packages/card-validator/types.ts
```diff
@@ -2,28 +2,34 @@ import { CardBrandName } from "types";
...
+  N extends string = string,
+  Local extends boolean = boolean
+> = {
+  name: N;
+  isLocal: Local;
   numberValidationRules: NumberValidationRules;
   securityCodeValidationRules: SecurityCodeValidationRules;
 };
 
-export type CardValidationOptions = {
```
joshpensky, line 20: same here re: removing types

maybe we just jsdoc it with `@deprecated`?

joshpensky (commented):
fantastic 🪄 just two little fixes!

packages/card-validator/types.ts
jmorgadosoares: Absolutely, I should have caught that. Fixing both asap

joshpensky (approved):
lovely work 😌  let's just hold off on merging until the deploy queue is fixed

PR #886, opened 2026-06-01
Add custom brand support to JS SDK

jmorgadosoares:
# Why

Adds the public API surface for custom card brands to the browser SDK and wires them through to the card iframe for validation.

Part of the Custom Brands RFC. Blocked on CARD-724.

Linear: CARD-725 (CARD-726, CARD-727, CARD-728)

# How

- Added `CustomBrand` and `BrandOptions` types to the shared `types` package, alongside a `CARD_BRAND_NAMES` constant used for brand type narrowing
- Added `brands.create(name, options)` to `EvervaultClient`, consistent with the `transactions.create` pattern. Always sets `isLocal: true` — custom brands are always local by design
- Added `customBrands?: CustomBrand[]` to `CardOptions` and `CardConfig`
- Wired `customBrands` through the postMessage layer (`EV_INIT` / `EV_UPDATE`) so the card iframe receives brand definitions before validation begins
- Passed `customBrands` into all `validateNumber` and `validateCVC` calls inside the iframe
- Updated `BrandIcon` to show a custom brand's `iconSrc` when matched, falling back to the default icon set
- Custom brands are always accepted regardless of `acceptedBrands` — `isBrandSupported` handles this by short-circuiting on any brand found in `localBrands` that matches a custom brand name

### DIFF
--- a/e2e-tests/ui-components/tests/customBrands.spec.js
+++ b/e2e-tests/ui-components/tests/customBrands.spec.js
@@ -152,6 +152,33 @@ test.describe("custom brands", () => {
     await expect.poll(async () => values.errors).toBeNull();
   });
 
+  test("custom brand iconSrc is shown when the card number matches", async ({
+    page,
+  }) => {
+    const ICON_SRC = "https://example.com/acme-icon.png";
+
+    await page.exposeFunction("handleChange", () => {});
+
+    await page.evaluate(
+      ({ customBrandOptions, iconSrc }) => {
+        const customBrand = window.evervault.brands.create("acme-card", {
+          ...customBrandOptions,
+          iconSrc,
+        });
+
+        const card = window.evervault.ui.card({ customBrands: [customBrand], icons: true });
+        card.on("change", window.handleChange);
+        card.mount("#form");
+      },
+      { customBrandOptions: CUSTOM_BRAND_OPTIONS, iconSrc: ICON_SRC }
+    );
+
+    const frame = page.frameLocator("iframe[data-evervault]");
+    await frame.getByLabel("Number").fill(CUSTOM_BRAND_CARD.number);
+
+    await expect(frame.locator(`img[src="${ICON_SRC}"]`)).toBeVisible();
+  });
+
   test("CVC is truncated to the maximum allowed length for the custom brand", async ({
     page,
   }) => {
--- a/e2e-tests/ui-components/tests/customBrands.spec.js
+++ b/e2e-tests/ui-components/tests/customBrands.spec.js
@@ -1,0 +1,153 @@
+import { test, expect, VALID_CARDS } from "../utils";
+
+// A custom brand used across tests.
+// Range [88000, 88999], 16 digits, Luhn-checked, 4-digit CVC.
+// Card number 8810000000000003 matches this range, is 16 digits, and passes Luhn.
+const CUSTOM_BRAND_CARD = {
+  number: "8810000000000003",
+  month: "01",
+  year: "35",
+  cvc: "1234",
+};
+
+function mountWithCustomBrands(page, opts = {}) {
+  return page.evaluate((opts) => {
+    const customBrand = window.evervault.brands.create("acme-card", {
+      numberValidationRules: {
+        luhnCheck: true,
+        ranges: [[88000, 88999]],
+        lengths: [16],
+      },
+      securityCodeValidationRules: {
+        lengths: [4],
+      },
+    });
+
+    const card = window.evervault.ui.card({
+      ...opts,
+      customBrands: [customBrand],
+    });
+    card.on("change", window.handleChange);
+    card.mount("#form");
+  }, opts);
+}
+
+test.describe("custom brands", () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto("http://localhost:4005");
+    await page.waitForFunction(() => window.Evervault);
+  });
+
+  test("custom brand is detected in localBrands and not in brand", async ({
+    page,
+  }) => {
+    let values = {};
+    await page.exposeFunction("handleChange", (newValues) => {
+      values = newValues;
+    });
+
+    await mountWithCustomBrands(page);
+
+    const frame = page.frameLocator("iframe[data-evervault]");
+    await frame.getByLabel("Number").fill(CUSTOM_BRAND_CARD.number);
+
+    await expect
+      .poll(async () => values.card?.localBrands)
+      .toContain("acme-card");
+    await expect.poll(async () => values.card?.brand).toBeNull();
+  });
+
+  test("custom brand card is valid with matching length, luhn, and CVC", async ({
+    page,
+  }) => {
+    let values = {};
+    await page.exposeFunction("handleChange", (newValues) => {
+      values = newValues;
+    });
+
+    await mountWithCustomBrands(page);
+
+    const frame = page.frameLocator("iframe[data-evervault]");
+    await frame.getByLabel("Number").fill(CUSTOM_BRAND_CARD.number);
+    await frame
+      .getByLabel("Expiration")
+      .fill(`${CUSTOM_BRAND_CARD.month}/${CUSTOM_BRAND_CARD.year}`);
+    await frame.getByLabel("CVC").fill(CUSTOM_BRAND_CARD.cvc);
+
+    await expect.poll(async () => values.isValid).toBeTruthy();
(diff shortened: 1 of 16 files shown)


packages/ui-components/src/Card/utilities.ts
```diff
@@ -141,9 +165,12 @@ async function encryptedNumber(ev: PromisifiedEvervaultClient, number: string) {
 async function encryptedCVC(
   ev: PromisifiedEvervaultClient,
   cvc: string,
-  cardNumber: string
+  cardNumber: string,
+  opts?: {
+    customBrands?: CustomBrand[];
+  }
```
joshpensky: **nit:** we should just pull the `CardNumberValidationOptions` type in here

packages/ui-components/src/Card/BrandIcon.tsx
```diff
@@ -1,19 +1,27 @@
...
-  let icon = (icons as Record<string, string>)[
-    brand ?? localBrands?.[0] ?? "default"
-  ];
-  icon = icon || icons.default;
+  const { brand, localBrands } = validateNumber(number, { customBrands });
+  const localBrand = localBrands[0];
+  const customBrand = customBrands?.find((b) => b.name === localBrand);
+  const brandName = brand ?? localBrand ?? "default";
+
+  let icon = customBrand?.iconSrc;
```
joshpensky: I think this would override a global brand match if a custom brand was also found?

joshpensky: Another way to do this is extend the `icons` map and keep the original logic:

```ts
const customIcons = Object.fromEntries(
  customBrands.map(b => [b.name, b.iconSrc])
)
const allIcons = { ...icons, ...customIcons }
```

packages/ui-components/src/Card/index.tsx
joshpensky, line 1: We should add some e2e tests here to confirm functionality!

joshpensky (commented):
nice work! 🧙 left a few comments

packages/ui-components/src/Card/BrandIcon.tsx
jmorgadosoares: > I think this would override a global brand match if a custom brand was also found?

It would, my gut was that clients would expect their custom brands to always override our default brands, no? 🤔 

> Another way to do this is extend the icons map and keep the original logic:

👍

joshpensky: Yeah that's a good point!

jmorgadosoares: Cool! Simplified the logic as suggested but made the order of priority `local brand > global brand > default` (local brands also match custom brands)

packages/ui-components/src/Card/BrandIcon.tsx
```diff
@@ -1,19 +1,25 @@
...
+  const { brand, localBrands } = validateNumber(number, { customBrands });
+  const customIcons = Object.fromEntries(
+    customBrands?.map((b) => [b.name, b.iconSrc]) ?? []
+  );
+  const allIcons: Record<string, string | undefined> = {
+    ...icons,
+    ...customIcons,
+  };
+  // local or custom brand > global brand > default icon
+  const icon = allIcons[localBrands[0] ?? brand ?? "default"] ?? icons.default;
```
joshpensky, line 22: On second thought, this actually is changing the current behavior that shows `global > local > default`

to prevent any unexpected changes, maybe we only bump custom brands in the priority order so: `custom > global > local > default`

thoughts?

e2e-tests/ui-components/tests/customBrands.spec.js
```diff
@@ -0,0 +1,153 @@
...
+  return page.evaluate((opts) => {
+    const customBrand = window.evervault.brands.create("acme-card", {
+      numberValidationRules: {
+        luhnCheck: true,
+        ranges: [[88000, 88999]],
+        lengths: [16],
+      },
+      securityCodeValidationRules: {
+        lengths: [4],
+      },
```
joshpensky: Let's also add a test for which icon should show (based on whatever we decide in that other comment)

packages/ui-components/src/Card/BrandIcon.tsx
jmorgadosoares: That is a good point, two counter arguments:
- imo `global > local > default` is an anti-pattern. If a user enables `elo` (local) and `visa` (global) - they have overlapping ranges - they would expect `elo` numbers to be shown as `elo`. The patern should be `specific > less specific > fallback`. So in practice this minor bump fixes an anti pattern here which becomes more relevant with the custom brand support
- on the preventing any unexpected changes point, all this changes is which icon is shown. Imo very low risk/impact so if you agree I'm comfortable with shipping this as a minor bump

e2e-tests/ui-components/tests/customBrands.spec.js
jmorgadosoares: Agreed!

packages/ui-components/src/Card/BrandIcon.tsx
joshpensky: Yeah okay, those are both fair points!

let's just flag this in Slack when you announce the changes just in case a customer notices the update 🫡

joshpensky (approved):
a great update 👏

packages/ui-components/src/Card/BrandIcon.tsx
jmorgadosoares: Will do!! 🫡

packages/ui-components/src/Card/CardNumber.tsx
```diff
@@ -55,29 +102,32 @@ export function CardNumber({
...
-      {
-        mask: "0000 0000 0000 0000 000",
-        brand: "unionpay",
-      },
-      {
-        mask: "0000 000000 00000",
-        brand: "american-express",
-      },
-    ] as CardMask[],
+    mask: masks as CardMask[],
```
joshpensky: **nit** I think we can remove this type assertion

```suggestion
    mask: masks,
```

packages/ui-components/src/Card/CardNumber.tsx
```diff
@@ -55,29 +102,32 @@ export function CardNumber({
...
+      }
 
-      const mask = dynamicMasked.compiledMasks.find((m) => {
+      const brandMask = dynamicMasked.compiledMasks.find((m) => {
         const maskBrand = (m as CardMask).brand;
         return maskBrand === brand;
       });
 
-      return mask ?? dynamicMasked.compiledMasks[0];
+      return brandMask ?? dynamicMasked.compiledMasks[0];
```
joshpensky: would love to pull this out into a unit-testable function since this also powers the `autoProgress` logic. maybe:

```tsx
function getDynamicMask(masks, number, { customBrands })
```

and it can be called here like:

```tsx
dispatch: (appended, dynamicMasked) => {
  const number = dynamicMasked.value + appended
  return getDynamicMask(dynamicMasked.compiledMasks, number, { customBrands })
}
```

packages/ui-components/src/Card/CardNumber.tsx
```diff
@@ -18,13 +19,33 @@ interface CardNumberProps {
...
 
+function cardMaskFromLength(length: number): string {
+  const parts = [];
+  let remaining = length;
+  while (remaining > 0) {
+    parts.push("0".repeat(Math.min(4, remaining)));
+    remaining -= 4;
+  }
+  return parts.join(" ");
+}
```
joshpensky: would love a simple unit test on this one!

packages/ui-components/src/Card/CardCVC.tsx
```diff
@@ -48,10 +51,23 @@ export const CardCVC = forwardRef<HTMLInputElement, CVCProps>(
...
+
+      const matchedCustomBrand = customBrands?.find((b) =>
+        localBrands.includes(b.name)
+      );
+      if (matchedCustomBrand) {
+        const maxLength = Math.max(
+          ...matchedCustomBrand.securityCodeValidationRules.lengths
+        );
+        return "0".repeat(maxLength);
+      }
```
joshpensky: same here with a simple unit test!

joshpensky (commented):
just requesting some small unit tests, but otherwise looks good! 🥂

joshpensky (approved):
Image

PR #888, opened 2026-06-02
Add custom brand support to React SDK

jmorgadosoares:
# Why
Part of the Custom Brands feature ([CARD-729](https://linear.app/evervault/issue/CARD-729/add-custom-brand-support-to-react-sdk)).

Exposes `customBrands` in the React SDK's `<Card>` component so consumers can pass custom card brand definitions without importing types from `@evervault/browser` directly.

# How
- Added `customBrands?: CustomBrand[]` prop to `CardProps` in `@evervault/react`
- Added `customBrands` to the `useMemo` config object and its dependency array so re-renders propagate changes correctly via `instance.update(config)`
- `BrandOptions` and `CustomBrand` types are re-exported from `@evervault/react` via the existing `export type * from "types"` — no extra changes needed
- Added changeset for a minor version bump

### DIFF
--- a/packages/browser/lib/main.ts
+++ b/packages/browser/lib/main.ts
@@ -14,12 +14,11 @@ import {
 import type { InputSettings, RevealSettings } from "./types";
 import { Transaction } from "./resources/transaction";
 import {
-  BrandOptions,
-  CustomBrand,
   CreateTransactionDetails,
   DisbursementTransactionDetails,
   RecurringTransactionDetails,
 } from "types";
+import { createBrand } from "shared/createBrand";
 
 export type * from "types";
 export type * from "./config";
@@ -333,20 +332,6 @@ export default class EvervaultClient {
   }
 
   get brands() {
-    return {
-      create: (name: string, options: BrandOptions): CustomBrand => ({
-        name,
-        isLocal: true,
-        numberValidationRules: {
-          luhnCheck: options.numberValidationRules.luhnCheck ?? true,
-          ranges: options.numberValidationRules.ranges,
-          lengths: options.numberValidationRules.lengths,
-        },
-        securityCodeValidationRules: {
-          lengths: options.securityCodeValidationRules.lengths,
-        },
-        iconSrc: options.iconSrc,
-      }),
-    };
+    return { create: createBrand };
   }
 }
--- a/packages/browser/lib/main.ts
+++ b/packages/browser/lib/main.ts
@@ -14,6 +14,7 @@ import {
 import type { InputSettings, RevealSettings } from "./types";
 import { Transaction } from "./resources/transaction";
 import {
+  BrandOptions,
   CreateTransactionDetails,
   DisbursementTransactionDetails,
   RecurringTransactionDetails,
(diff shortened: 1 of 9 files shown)


packages/react/lib/load.ts
```diff
@@ -14,6 +16,24 @@ export class PromisifiedEvervaultClient extends Promise<EvervaultClient> {
...
+          ranges: options.numberValidationRules.ranges,
+          lengths: options.numberValidationRules.lengths,
+        },
+        securityCodeValidationRules: {
+          lengths: options.securityCodeValidationRules.lengths,
+        },
+        iconSrc: options.iconSrc,
+      }),
+    };
+  }
```
joshpensky, line 25: can we move this logic to a shared place so both the browser and React libs can pull from it?

ideally we could've just called `ev.brands.create` but it's async in this case 😓

joshpensky (commented):
super simple 😎 just one comment

packages/react/lib/load.ts
jmorgadosoares: Yeah, I forgot to leave a comment on this. I didn't wanna export `createBrand ` from `@evervault/browser` to reuse in react because it'd expose the functions hence the repeated logic. I hadn't seen the shared package, I'll move the logic there

On the `async` situation, yeah it's not ideal for this specific use case but this way it's clean enough to use inside a `useEffect`

packages/browser/lib/main.ts
```diff
@@ -14,12 +14,11 @@ import {
 import type { InputSettings, RevealSettings } from "./types";
 import { Transaction } from "./resources/transaction";
 import {
-  BrandOptions,
-  CustomBrand,
   CreateTransactionDetails,
   DisbursementTransactionDetails,
   RecurringTransactionDetails,
 } from "types";
+import { createBrand } from "shared";
```
joshpensky: Nice call on using `shared`! Just a note there: we recently ran into something where our barrel export in `shared` was also pulling React into the browser lib when it wasn't needed 😅

Max added a fix in https://github.com/evervault/evervault-js/pull/871 where we add an entry to the package `exports` for importing that specific module

packages/react/lib/load.ts
```diff
@@ -1,6 +1,7 @@
 import EvervaultClient, {
   CustomConfig as BrowserConfig,
 } from "@evervault/browser";
+import { createBrand } from "shared";
```
joshpensky, line 5: re: other comment, should be good to just use `shared` here

joshpensky (approved):
awesome ⚛️

packages/react/lib/load.ts
```diff
@@ -14,6 +16,24 @@ export class PromisifiedEvervaultClient extends Promise<EvervaultClient> {
...
+  get transactions() {
+    return {
+      create: async (
+        ...args: Parameters<EvervaultClient["transactions"]["create"]>
+      ) => {
+        const ev = await this;
+        return ev.transactions.create(...args);
+      },
+    };
+  }
```
joshpensky: let's leave this one off for now, it's only used for Apple Pay which isn't supported in React yet

joshpensky (commented):
very nice! just one small change, but otherwise gtg 🎸

joshpensky (approved):
🚢

PR #889, opened 2026-06-02
feat: allow optional cvc

oisincurran:
Creates a new `validation.cvc.optional` option. If enabled, an empty field passes validation, otherwise regular CVC validation is enforced.


joshpensky (approved):
sick! 💥

PR #892, opened 2026-06-02
fix: split out aws operations from preflight acceptance testing in playwright container

lfarrel6:
# Why

preview deployment aren't working as expected due to the lack of access to an aws cli. This isn't vendored into the playwright container that the preview deployment step is now running in. 

These 2 steps should happen separately - we should push the artifact from the runner and then spawn the playwright image to verify it.

# How

Split out artifact deployment and testing.

### DIFF
--- a/.github/workflows/publish-static-bundle.yml
+++ b/.github/workflows/publish-static-bundle.yml
@@ -66,8 +66,6 @@ jobs:
   # This is to ensure that the preview deployment is working as expected before we promote to production.
   deploy-preview:
     runs-on: ubuntu-latest
-    container:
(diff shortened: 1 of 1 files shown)


lfarrel6:
@SocketSecurity ignore npm/@react-native/debugger-frontend@0.76.9

Flagged for obfuscated code - report is flagging 11 base64 encoded strings in a test. The base64 encoded strings are encoded X509 certificates.

CianT123 (approved):

matt-evervault (approved):

joshpensky (approved):
thank you 💜

PR #899, opened 2026-06-09
Fix React SDK loading strategy

joshpensky:
# Why

Our old React SDK loading strategy was not working for a few reasons:
- The in-memory map didn't account for multiple providers or existing scripts loaded
- The onLoadError callback wasn't called when the script failed to load

This new loading strategy works like so:
1. The SDK will try to inject the script if one is not already present
2. If the script loading fails, we will call `onLoadError` with the error that occurred
3. If the `ref.reload()` function is called, we will inject a new script to attempt reloading
    - This will include a new `?attempt=` query param, which will force it to refetch

https://github.com/user-attachments/assets/a9d183e7-f653-4e13-8816-b542420f62e4

# How

- Separate script loading helpers into different files
- Add unit tests to all helpers to ensure scripts load as expected
- Update React example to include way to test loading failures


jmorgadosoares (approved):
Look good!

PR #907, opened 2026-06-17
Fix domain resolution for Apple Pay SDK

ana-maksimovskikh:
# Why
When the @evervault/browser Apple Pay component runs inside a cross-origin iframe, the merchant session was issued with domainName set to the iframe's own origin rather than the top-level page origin. 
Safari validates a cross-origin-iframe Apple Pay session against the top-level origin, so completeMerchantValidation was rejected and the sheet showed "Payment Not Completed" after the user authorized.

The cause was in the Transaction constructor, which derived domain unconditionally from window.location.origin — i.e. the current frame.

# Fix
Resolve the validation domain from the top-level document origin via window.location.ancestorOrigins:

New resolveTopLevelDomain() helper returns the top-level origin (the furthest ancestorOrigins entry — the address-bar origin, regardless of nesting depth). ancestorOrigins is browser-populated and cannot be spoofed by page JS, so it's a safe source of truth.

Falls back to the current frame's origin when not embedded, when ancestorOrigins is unavailable, or when the top origin is opaque ("null").

The existing backend check (which validates the supplied domain against the merchant's registered, active Apple Pay domains) remains the authoritative trust boundary, so no new client-trust surface is introduced.

# Tests

- Tested manually:
  - Reproduced the issue (ApplePay doesn't work inside iframe)
  - With this fix applied, ApplePay works both with/without iframe

### DIFF
--- a/packages/browser/lib/utils/resolveTopLevelDomain.ts
+++ b/packages/browser/lib/utils/resolveTopLevelDomain.ts
@@ -1,0 +1,37 @@
+function stripProtocol(origin: string): string {
+  return origin.replace(/https?:\/\//, "");
+}
+
+/**
+ * Resolves the domain used for Apple Pay merchant validation.
+ *
+ * Apple Pay (and Safari) validate a merchant session against the *top-level*
+ * document's origin — the one in the address bar that the shopper actually
+ * sees. When the SDK runs inside a cross-origin iframe (e.g. a PSP-hosted
+ * widget embedded in a merchant store), `window.location.origin` is the
+ * iframe's own origin, not the top-level page, so the merchant session is
+ * issued for the wrong domain and `completeMerchantValidation` is rejected.
+ *
+ * `window.location.ancestorOrigins` is populated by the browser from the real
+ * frame tree and cannot be spoofed by page-level JS, so it is the only
+ * trustworthy source for the top-level origin. It is ordered from the
+ * immediate parent (index 0) to the top-level document (last index), so the
+ * final entry is the address-bar origin regardless of nesting depth.
+ *
+ * Falls back to the current frame's origin when not embedded, when
+ * `ancestorOrigins` is unavailable, or when the top-level origin is opaque
+ * (reported as "null", e.g. a sandboxed ancestor) — an opaque origin can never
+ * be a registered Apple Pay domain, so we avoid sending "null" downstream.
+ */
+export default function resolveTopLevelDomain(): string {
+  const { ancestorOrigins } = window.location;
+
+  if (ancestorOrigins && ancestorOrigins.length > 0) {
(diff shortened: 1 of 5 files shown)


jakekgrog (approved):
Nice work!

packages/browser/lib/utils/resolveTopLevelDomain.ts
```diff
@@ -0,0 +1,37 @@
...
+ * `ancestorOrigins` is unavailable, or when the top-level origin is opaque
+ * (reported as "null", e.g. a sandboxed ancestor) — an opaque origin can never
+ * be a registered Apple Pay domain, so we avoid sending "null" downstream.
+ */
+export default function resolveTopLevelDomain(): string {
+  const { ancestorOrigins } = window.location;
+
+  if (ancestorOrigins && ancestorOrigins.length > 0) {
+    const topLevelOrigin = ancestorOrigins[ancestorOrigins.length - 1];
+    if (topLevelOrigin && topLevelOrigin !== "null") {
```
joshpensky, line 31: Should we only be checking the very last ancestor? like if the array looks like `["https://hello.com", "null"]`, should we fallback to the second-to-last ancestor, or just default to iframe origin then?

I'm not super familiar with this area so the question may not be relevant here (lol)

joshpensky (approved):
Nice! Added one question, but non-blocking

packages/browser/lib/utils/resolveTopLevelDomain.ts
ana-maksimovskikh: Good question!

Short version: we intentionally only use the last entry (the top-level origin) and don't walk back to intermediate ancestors. It comes down to what Apple Pay actually validates against. 

Safari validates the merchant session against the top-level document's origin — the one in the address bar — not any intermediate frame. In `["https://hello.com", "null"]` example, the top-level is the "null" (opaque) one, and there's no real registered domain to validate against — the payment can't succeed regardless of what we send.

Falling back to the second-to-last ancestor (hello.com) wouldn't fix that: Safari would still be validating against the opaque top-level, so it'd fail anyway — and we'd be telling Apple the payment is for a domain the user isn't actually on, which is exactly the origin-confusion the top-level binding is meant to prevent. So walking the chain is a security smell with no upside.

The "null" → iframe-origin fallback isn't really trying to "recover" — an opaque top-level means Apple Pay fails downstream either way. It just avoids sending the literal string "null" and keeps the prior behaviour for that degenerate case. The backend is the real backstop here: it only mints a session for a domain that's registered + active for the merchant, so a wrong value gets rejected there too.

And it's a rare case in practice — a top-level origin is only "null" for things like a sandboxed (without allow-same-origin) or data: top document, not a normal merchant page

PR #909, opened 2026-06-19
add name field to CardTranslations with errors.regex support

jmorgadosoares:
# Why

`CardTranslations` in `packages/types/uiComponents.ts` was missing an explicit `name` field, so `translations.name.errors.regex` had no type — even though the runtime default translations in `packages/ui-components/src/Card/translations.ts` already include `name.errors.regex`. Fixes EXP-1023.

# How

Added `name?: CardFieldTranslations<{ invalid?: string; regex?: string }>` to the `CardTranslations` interface, consistent with the existing `DEFAULT_TRANSLATIONS` shape.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FCmCRYRgLkTrepYJ4Lp7K5)_

### DIFF
--- a/packages/types/uiComponents.ts
+++ b/packages/types/uiComponents.ts
@@ -115,6 +115,7 @@ export interface CardTranslations extends TranslationsObject {
   }>;
   expiry: CardFieldTranslations<{ invalid?: string }>;
   cvc: CardFieldTranslations<{ invalid?: string }>;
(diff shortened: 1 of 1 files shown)


jmorgadosoares:
The type of translations is an extension of TranslationsObject which is [key: string]: string | TranslationsObject | undefined;. This means this change is redundant. Closing the PR

PR #919, opened 2026-06-24
Add optional appleMerchantId to Apple Pay web SDK (CARD-851)

Jason-Papa:
Expose a custom Apple merchant identifier option while defaulting to merchant.com.evervault.{merchantId} for Evervault domain verification.


Jason-Papa:
Managed to pass an invalid merchant id which resulted in "unavailable" state. Then with the default one (registered merchant) it is working, and finally it worked with a different merchant that was registered to the same domain. Setting a different merchant with a different domain, and using that merchant resulted in "unavailable".

alexander-rw (approved):

PR #923, opened 2026-06-29
Support React Native 0.79 and Expo 53 in @evervault/react-native

joshpensky:
# Why

In order to upgrade our Android dependency, we will need to start testing against slightly newer versions of React Native and Expo. This also required us to update our React dev dependency to 19, which surfaced some type and test errors

# How

- Upgrades React Native (and related deps) to 0.79
- Upgrades Expo (and related deps) to SDK 53
- Upgrades React (and related deps) to 19
- Fixes type errors caused by React 19 update
- Fixes React Native tests caused by React 19 update


packages/react-native-v2/package.json
```diff
@@ -31,31 +31,37 @@
...
+    "expo": ">=53.0.0",
     "react": ">=18.0.0",
     "react-native": ">=0.76.0",
     "react-native-webview": ">=13.0.0"
   },
+  "peerDependenciesMeta": {
+    "expo": {
+      "optional": true
+    }
+  },
```
joshpensky, line 48: Expo is not a required peerDep of our RN library, but if it is installed we should warn that we require 53+

matt-evervault (approved):

PR #926, opened 2026-07-06
Upgrade React Native iOS `EvervaultCore` dependency

joshpensky:
# Why

This PR upgrades the React Native SDK's iOS `EvervaultCore` SDK dependency from 0.3.4 → 2.1.0

## Background

The used version 0.3.4 was published to Cocoapods years ago and is no longer actively maintained by us. The current version 2.1.0 is the latest published SDK, but it's only available on Swift Package Manager (SPM).

The first idea was to also publish the iOS SDK to Cocoapods. However, Cocoapods trunk (where all pods are currently published to) is being deprecated and will turn read-only by year end. We could still continue with the Cocoapods route but we would need to publish to our own Cocoapods repo, which would require additional configuration in every customer's app to declare the other Cocoapods repo.

The second idea was to import the dependency directly from SPM. However, there is currently a bug when using SPM with Expo that causes duplicate symbol errors: https://github.com/expo/expo/issues/37813. The current recommended "fix" is to force enable dynamic linking, but that would also require additional configuration that may not work with other React Native dependencies (off the top of my head I could think of Firebase which requires static linking).

Which lands us here: vendoring the iOS SDK source directly into our React Native library. On prebuild/prepack, we will fetch the latest release tag and store it in the `ios/vendor` directory. This will ensure we still have access to the latest SDK.

## Future

Ideally in the future React Native & Expo will fully support SPM packages and we don't need this workaround.

If we need to pull in `EvervaultInputs` for example, we can simply append the dependency to our package.json list:

```diff
 "swiftDependencies": {
   "evervault-ios": [
     "EvervaultCore",
+    "EvervaultInputs"
   ]
 }
```

# How

- Add new `vendor:ios` command script, which copies release source to the RN library `ios/vendor` directory


packages/react-native-v2/ios/NativeEvervault.swift
```diff
@@ -1,5 +1,3 @@
-import Evervault
```
joshpensky, line 1: `import` no longer needed since source is bundled in with the SDK

packages/react-native-v2/package.json
```diff
@@ -63,20 +73,22 @@
     "tsconfig": "workspace:*",
     "vitest": "catalog:"
   },
+  "ios": {
+    "spmDependencies": [
+      {
+        "name": "EvervaultCore",
+        "repo": "evervault/evervault-ios",
+        "version": "2.1.0",
```
joshpensky: Would love a way to pull this into dependabot somehow and ensure we're always on the latest version

packages/react-native-v2/scripts/vendor-spm-dependencies.mjs
```diff
@@ -0,0 +1,55 @@
...
+
+  const repoName = repo.split("/").pop();
+  const workDir = mkdtempSync(join(tmpdir(), `${repoName}-`));
+  const tarballPath = join(workDir, `${repoName}.tar.gz`);
+  
+  try {
+    // Download & extract the tarball
+    execFileSync("curl", [
+      "-sL",
+      `https://github.com/${repo}/archive/refs/tags/${version}.tar.gz`,
```
matt-evervault: Kind of hacky suggestion, but you would get dependabot updates, is to create a Package.swift in the root (or under the react-native target) that has no targets, but depends on the Evervault SDK.  This scrip then takes the version from the Package.lock that committed to the repo, instead of from the `package.json`, so you can use existing tools to infer the version and keep it up to date :)  Give me a shout if needed on this!

packages/react-native-v2/package.json
matt-evervault: See comment above :)

packages/react-native-v2/scripts/vendor-spm-dependencies.mjs
joshpensky: This worked great!

I added a `Package.swift` (which resolves to `Package.resolved`) and using that to pin versions. Then `package.json` is just used to choose sources to copy over

I also updated `dependabot.yml` to check that directory for swift versions

matt-evervault (approved):

PR #927, opened 2026-07-07
Surface CardEnrichment data & phonetic name for ApplePay

ana-maksimovskikh:
https://linear.app/evervault/issue/CARD-872/b-result-surface-parity-apple-pay-ios-web

### DIFF
--- a/.changeset/apple-pay-card-enrichment-phonetic-name.md
+++ b/.changeset/apple-pay-card-enrichment-phonetic-name.md
@@ -1,0 +1,6 @@
+---
+"@evervault/browser": minor
+"@evervault/js": minor
+---
+
+Surface Apple Pay card enrichment (`funding`, `segment`, `country`, `currency`, `issuer`) and phonetic name fields (`phoneticGivenName`, `phoneticFamilyName`) on `billingContact`/`shippingContact` in the `process()` payload. These fields were already sent by the backend and browser but were not previously exposed on the `EncryptedApplePayData` type.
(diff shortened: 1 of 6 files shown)


ana-maksimovskikh:
@joshpensky 

Hi Josh! I've added you case the PR contains fix for a newly introduced issue - just to be on the safe side 🙃 
Happy to split it into a separate PR if that's what you'd prefer


>Commit 8ea12959 — "Support React Native 0.79 and Expo 53 in @evervault/react-native ([#923](https://github.com/evervault/oxide/issues/923))", by Josh Pensky, merged to master Mon Jul 6, 2026, 11:28:22 +0100 (yesterday). That PR bumped the @types/react catalog version in pnpm-workspace.yaml:
```
-  "@types/react": ^18.2.37
+  "@types/react": ~19.0.14
````
>React 19's type definitions tightened useRef's overloads. Under React 18's types, calling useRef<T>() with no argument at all was fine — it returned a mutable MutableRefObject<T | undefined>. React 19 no longer allows that no-argument form for a type like this, so the same commit had to patch these exact call sites to pass an explicit initial value:
```
- const validationCallback = useRef<ValidationCallback<T>>();
+ const validationCallback = useRef<ValidationCallback<T>>(null);
- const frame = useRef<HTMLIFrameElement>();
+ const frame = useRef<HTMLIFrameElement>(null);
- const mask = useRef<ReturnType<typeof IMask>>();
+ const mask = useRef<ReturnType<typeof IMask>>(null);
```
>That fix silenced the "no argument" error, but it was subtly incomplete — passing null without also widening the generic type argument to include | null makes React 19's overload resolution pick the read-only RefObject<T> variant instead of the mutable one, which is exactly the TS2540: Cannot assign to 'current' error on push.

alexander-rw (approved):

joshpensky (approved):
Thanks for fixing this! 🙌

PR #947, opened 2026-07-21
Jason/card 1045/ap add coupon support

Jason-Papa:
# Why
Add coupon handling as well as fixing the availability function (due to a missing import)

PR #949, opened 2026-07-21
Jason/card 1046/add billing contact

Jason-Papa:
(to be merged after https://github.com/evervault/evervault-js/pull/947)
# Why
add prefil billing contact handling

### DIFF
--- a/examples/apple-pay/index.html
+++ b/examples/apple-pay/index.html
@@ -3,27 +3,25 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Apple Pay — contact prefill</title>
+    <title>Apple Pay — coupon + contact prefill</title>
   </head>
 
   <body>
     <div class="card">
-      <h1 class="title">Contact prefill test</h1>
+      <h1 class="title">Apple Pay test</h1>
       <p class="subtext">
         Open in <strong>Safari</strong> with a card in Wallet. Tap Apple Pay and
-        confirm the sheet shows:
+        confirm:
       </p>
       <ul class="subtext" style="margin-top: 0; padding-left: 1.25rem">
         <li>
-          Billing: <strong>Evervault Billing</strong> —
-          1 Prefill Billing St, Cupertino CA 95014
+          <strong>Contact prefill</strong> — billing
+          <em>Evervault Billing / 1 Prefill Billing St</em>, shipping
+          <em>Evervault Shipping / 2 Prefill Shipping Ave</em> (+ email/phone)
         </li>
         <li>
-          Shipping: <strong>Evervault Shipping</strong> —
-          2 Prefill Shipping Ave, plus email/phone
-        </li>
-        <li>
-          Optional: coupon <strong>SAVE20</strong> drops total $10 → $8
+          <strong>Coupon</strong> — enter <strong>SAVE20</strong>; total drops
+          $10.00 → $8.00
         </li>
       </ul>
       <div id="apple-pay"></div>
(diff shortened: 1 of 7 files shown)


packages/browser/lib/ui/ApplePay/index.ts
```diff
@@ -66,6 +67,18 @@ export type ApplePayButtonOptions = {
    * coupon message in the native Apple Pay UI.
    */
   onCouponCodeChange?: (couponCode: string) => Promise<CouponCodeChangeResult>;
+  /**
+   * Prefill billing contact on the Apple Pay sheet (ApplePayRequest.billingContact).
+   * Apple requires `requestBillingAddress: true` when the contact includes a postal
+   * address; this SDK does not enforce that pairing.
+   */
+  billingContact?: PaymentContact;
```
ana-maksimovskikh, line 78: Maybe something like?

```
/**
 * Prefill billing contact on the Apple Pay sheet (ApplePayRequest.billingContact).
 * Requires `requestBillingAddress: true` for the postal address to actually show.
 *
 * Payment and recurring transactions only. Apple Pay disbursements use a
 * separate recipient-contact model (`requiredRecipientDetails` on the
 * transaction) — this field has no effect on disbursement-type transactions.
 */
```

packages/browser/lib/ui/ApplePay/utilities.ts
```diff
@@ -477,19 +493,22 @@ function buildDisbursementSession(
...
+    version: 3,
+    merchantIdentifier: resolveMerchantIdentifier(
+      config.transaction.merchantId,
+      config.appleMerchantId
+    ),
+    merchantCapabilities,
+    supportedNetworks: config.allowedCardNetworks,
+    countryCode: tx.country,
+  };
+  applyContactFields(disbursementMethodData, config);
```
ana-maksimovskikh: Do we need this line 👀 

IIUC Apple Pay disbursements have no billing/shipping contact concept — recipient info is collected separately via `disbursementRequest.requiredRecipientContactFields` below.  
And any billingContact/shippingContact passed in config is inert here; harmless (unread extra keys) but has no effect on the sheet.

Maybe I got something wrong here though 🤔

packages/browser/lib/ui/ApplePay/index.ts
```diff
@@ -66,6 +67,18 @@ export type ApplePayButtonOptions = {
...
+  /**
+   * Prefill shipping contact on the Apple Pay sheet (ApplePayRequest.shippingContact).
+   * Apple requires `requestShipping: true` when the contact includes a postal
+   * address; this SDK does not enforce that pairing.
+   */
+  shippingContact?: PaymentContact;
```
ana-maksimovskikh, line 87: Same suggestion here:

```
/**
 * Prefill shipping contact on the Apple Pay sheet (ApplePayRequest.shippingContact).
 * Requires `requestShipping: true` for the postal address to actually show.
 *
 * Payment and recurring transactions only. Apple Pay disbursements use a
 * separate recipient-contact model (`requiredRecipientDetails` on the
 * transaction) — this field has no effect on disbursement-type transactions.
 */
```

packages/browser/lib/ui/ApplePay/utilities.ts
ana-maksimovskikh: If that's correct, but we want to keep this line for symmetry - maybe worth a comment

examples/apple-pay/index.html
```diff
@@ -3,19 +3,31 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Apple Pay — coupon test</title>
+    <title>Apple Pay — contact prefill</title>
```
ana-maksimovskikh: optional: should we consider widening the example vs overwriting it?

ana-maksimovskikh (approved):
Thank you @Jason-Papa 🔥 

All looks good - just some optional nits/suggestions, nothing blocking

PR #952, opened 2026-07-23
Prevent masked input flicker on React Native

joshpensky:
# Why

Since we know the length of the masked input value ahead of time, we can add the `maxLength` prop to our masked inputs to prevent flicker

## Before

https://github.com/user-attachments/assets/2cb3b14e-a49f-49cc-9079-f5029be7ed53

## After

https://github.com/user-attachments/assets/92dbbd79-54b6-48fb-999d-8af9d922a172

### DIFF
--- a/examples/expo/app.config.js
+++ b/examples/expo/app.config.js
@@ -60,6 +60,9 @@ module.exports = {
       },
       package: "com.evervault.expoexample",
     },
+    experiments: {
+      reactCanary: true,
+    },
     plugins: [
       [
         "expo-splash-screen",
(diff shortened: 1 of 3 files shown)


examples/expo/app.config.js
```diff
@@ -60,6 +60,9 @@ module.exports = {
       },
       package: "com.evervault.expoexample",
     },
+    experiments: {
+      reactCanary: true,
+    },
```
joshpensky, line 65: Unexpected side effect of [bumping React to 19.2.7](https://github.com/evervault/evervault-js/pull/932)

React Native 0.79/Expo 53 is pinned to work with only 19.0.0, so need to enable this experiment so it works with future 19 versions: https://docs.expo.dev/versions/latest/config/app/#reactcanary

matt-evervault (approved):

PR #953, opened 2026-07-23
Fix ApplePay checkout failure - desktop Chrome QR 

ana-maksimovskikh:
# Why
https://linear.app/evervault/issue/PE-112/fix-apple-pay-desktop-chrome-qr-checkout-failure

`event.target` is normally the reference back to the object that dispatched the event  — for a real `PaymentRequestUpdateEvent`, that's supposed to be the `PaymentRequest` instance itself, which is why the code expects `target.shippingAddress to work (the spec puts the current shipping address on the request object during this event).

In this bug, `event.target` is null instead of that instance. We know this specifically because of the exact wording of the error: Cannot read properties of null (reading 'shippingAddress'). V8 distinguishes null from undefined in that message, so the value being dereferenced is literally null — not a `PaymentRequest` missing a `shippingAddress` field, and not undefined.

Why would it be null at all? Our code never constructs this event — Apple's apple-pay-sdk.js polyfill In the normal same-device flow it presumably sets target correctly. In the desktop-Chrome + phone-QR "remote continuity" flow the session is reconstructed from a signal relayed from the phone, and Apple's polyfill apparently synthesizes the event without wiring up target at all, leaving it null.

**Tested** manually + unit test added

### DIFF
--- a/.changeset/shy-donuts-relax.md
+++ b/.changeset/shy-donuts-relax.md
@@ -1,0 +1,6 @@
+---
+"@evervault/browser": patch
+"@evervault/js": patch
+---
+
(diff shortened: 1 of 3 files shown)


Jason-Papa (approved):
lgtm

PR #954, opened 2026-07-23
Fix duplicate 3DS iframe posts and Evervault client memoization

jmorgadosoares:
# Why

Fixes a duplicate-submission bug in the 3DS challenge/browser-fingerprint iframes, flagged from [PE-107](https://linear.app/evervault/issue/PE-107): a customer session where the ACS rejected a CReq as already processed, causing the SDK to emit a `failure` even though the session went on to succeed. Root cause: `ChallengeFrame`/`BrowserFingerprint` guarded against re-posting to the iframe using "has a node ever been attached to this ref," not "have we already posted this specific CReq/method data." Under certain re-render conditions React detaches and reattaches a callback ref to the *same* iframe node, which reset that guard and let a second POST go out. Not yet confirmed as the exact cause of the PE-107 session (no duplicate CReq was visible in that customer's logs, only a successful transaction with a failed CRes), but it's a real, reproducible bug worth closing regardless.

Also includes some unrelated cleanup found while investigating:
- `useEvervaultClient`'s memoized client could be constructed twice for the same inputs if the memo factory ran more than once (its side effect, script injection, wasn't idempotent).
- A `cvv`/`cvc` field mismatch and a filename typo in the `three-d-secure`/`next-3ds` examples, which were out of date and made this harder to debug against.

# How

- **`ChallengeFrame.tsx` / `BrowserFingerprint.tsx`**: replaced the "have we ever attached a node" guard with a guard that tracks the actual content already posted (`postedCreq`/`postedData`), so a repeat ref invocation with the same node is a no-op, but genuinely new challenge data still posts.
- **`use-evervault-client.ts`**: cache the constructed client by dependency identity so a repeat factory invocation for the same deps reuses the existing client instead of constructing (and script-loading) a second one; read `onLoadError` via a ref so an inline callback prop doesn't itself invalidate the cache.
- **`ThreeDSecure.tsx`**: memoized `onMount`/`onUpdate` with `useCallback` so they don't change identity every render.
- **Examples**: fixed the `cvv`/`cvc` mismatch and `actons.server.ts` typo, and added `loadTimeout`/`isScriptLoadError` and a 3DS `error` event handler to demonstrate the newer error-handling APIs.


jakekgrog (approved):
Lgtm! Thanks for the time you've put into this so far

PR #963, opened 2026-07-28
add config passthrough

Jason-Papa:
# Why
add Apple Pay request-config passthrough

### DIFF
--- a/packages/browser/lib/ui/ApplePay/index.ts
+++ b/packages/browser/lib/ui/ApplePay/index.ts
@@ -85,6 +85,16 @@ export type ApplePayButtonOptions = {
    * transaction) — this field has no effect on disbursement-type transactions.
    */
   shippingContact?: PaymentContact;
+  /**
+   * Opaque merchant data included on the Apple Pay request
+   * (ApplePayRequest.applicationData). Must be a Base64-encoded string.
+   */
+  applicationData?: string;
+  /**
+   * Restrict payments to cards issued in these ISO 3166 country codes
+   * (ApplePayRequest.supportedCountries).
+   */
+  supportedCountries?: string[];
   prepareTransaction?: () => Promise<{
     amount?: number;
     lineItems?: TransactionLineItem[];
@@ -183,6 +193,8 @@ export default class ApplePayButton {
         onCouponCodeChange: this.#options.onCouponCodeChange,
         billingContact: this.#options.billingContact,
         shippingContact: this.#options.shippingContact,
+        applicationData: this.#options.applicationData,
+        supportedCountries: this.#options.supportedCountries,
         prepareTransaction: this.#options.prepareTransaction,
         appleMerchantId: this.#options.appleMerchantId,
       });
--- a/packages/browser/lib/ui/ApplePay/utilities.ts
+++ b/packages/browser/lib/ui/ApplePay/utilities.ts
@@ -25,6 +25,14 @@ import {
 import ApplePayButton from ".";
 import { RecurringPaymentIntervalUnit } from "types/uiComponents";
 
+/**
+ * Probe ceiling for `ApplePaySession.supportsVersion()`.
+ *
+ * Keep in sync with the major version of `@types/applepayjs` — a unit test fails
+ * if that package moves ahead of this constant.
+ */
+export const APPLE_PAY_MAX_VERSION = 14;
+
 type BuildSessionOptions = {
   transaction: TransactionDetailsWithDomain;
   allowedCardNetworks?: ApplePayCardNetwork[];
@@ -49,6 +57,8 @@ type BuildSessionOptions = {
   onCouponCodeChange?: (couponCode: string) => Promise<CouponCodeChangeResult>;
   billingContact?: PaymentContact;
   shippingContact?: PaymentContact;
+  applicationData?: string;
+  supportedCountries?: string[];
   prepareTransaction?: () => Promise<{
     amount?: number;
     lineItems?: TransactionLineItem[];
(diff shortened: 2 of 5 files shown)


packages/browser/test/applePay.test.ts
```diff
@@ -676,6 +684,170 @@ describe("buildSession contact prefill", () => {
...
+        amount: { value: "0.80", currency: "USD" },
+        pending: true,
+      },
+    ]);
+  });
+
+  it("defaults Apple Pay version to 3 when ApplePaySession is unavailable", async () => {
+    await buildSession(applePay, { transaction });
+    expect(paymentMethodDataCalls[0].version).toBe(3);
+  });
```
ana-maksimovskikh, line 840: Only `buildSession`-level version check covers the fallback-to-3 case. 
Should we add another one that proves a *resolved, non-default* version actually propagates through `buildApplePayMethodData` for both the payment and disbursement paths?

It might silently break if those two method-data builders are ever refactored apart again.

```suggestion
  });

  it("propagates a resolved non-default version to payment and disbursement method data", async () => {
    const originalApplePaySession = (
      globalThis as unknown as { ApplePaySession?: unknown }
    ).ApplePaySession;
    (globalThis as unknown as { ApplePaySession: unknown }).ApplePaySession = {
      supportsVersion: (version: number) => version <= 9,
    };

    try {
      await buildSession(applePay, { transaction });
      expect(paymentMethodDataCalls[0].version).toBe(9);

      paymentMethodDataCalls.length = 0;

      await buildSession(applePay, { transaction: disbursementTransaction });
      expect(paymentMethodDataCalls[0].version).toBe(9);
    } finally {
      if (originalApplePaySession === undefined) {
        delete (globalThis as unknown as { ApplePaySession?: unknown })
          .ApplePaySession;
      } else {
        (globalThis as unknown as { ApplePaySession: unknown }).ApplePaySession =
          originalApplePaySession;
      }
    }
  });
```

packages/browser/test/applePay.test.ts
```diff
@@ -676,6 +684,170 @@ describe("buildSession contact prefill", () => {
...
+
+    expect(resolveApplePayVersion(14)).toBe(3);
+  });
+
+  it("falls back to 3 when ApplePaySession is missing", () => {
+    delete (globalThis as unknown as { ApplePaySession?: unknown })
+      .ApplePaySession;
+
+    expect(resolveApplePayVersion()).toBe(3);
+  });
```
ana-maksimovskikh, line 918: Low-risk, but cheap to close off

```suggestion
  });

  it("falls back to 3 when supportsVersion is not a function", () => {
    (globalThis as unknown as { ApplePaySession: unknown }).ApplePaySession = {};

    expect(resolveApplePayVersion(14)).toBe(3);
  });
```

ana-maksimovskikh (commented):
Two small test-coverage suggestions for the `supportsVersion` gating (CARD-1047) — nice work overall, this closes out the batch cleanly.

packages/browser/lib/ui/ApplePay/utilities.ts
```diff
@@ -492,25 +534,29 @@ function buildDisbursementSession(
...
+    version: resolveApplePayVersion(),
+    merchantIdentifier: resolveMerchantIdentifier(
+      config.transaction.merchantId,
+      config.appleMerchantId
+    ),
+    merchantCapabilities,
+    supportedNetworks: config.allowedCardNetworks,
+    countryCode: tx.country,
+  };
+  applyRequestPassthroughFields(disbursementMethodData, config);
```
ana-maksimovskikh, line 566: Optional: this block duplicates `buildApplePayMethodData` - consider parameterizing and reusing it

packages/browser/lib/ui/ApplePay/index.ts
```diff
@@ -85,6 +85,16 @@ export type ApplePayButtonOptions = {
    * transaction) — this field has no effect on disbursement-type transactions.
    */
   shippingContact?: PaymentContact;
+  /**
+   * Opaque merchant data included on the Apple Pay request
+   * (ApplePayRequest.applicationData). Must be a Base64-encoded string.
```
ana-maksimovskikh, line 90: Do we have a validation for this somewhere?

packages/browser/lib/ui/ApplePay/utilities.ts
```diff
@@ -408,14 +457,7 @@ function buildRecurringSession(
   tx: RecurringTransactionDetails
 ) {
   console.log("Building recurring session");
```
ana-maksimovskikh: nit: might be a leftover debug print from before

packages/browser/test/applePay.test.ts
```diff
@@ -676,6 +684,170 @@ describe("buildSession contact prefill", () => {
...
+        label: "Item",
+        amount: { value: "10.00", currency: "USD" },
+      },
+      {
+        label: "Estimated tax",
+        amount: { value: "0.80", currency: "USD" },
+        pending: true,
+      },
+    ]);
+  });
```
ana-maksimovskikh, line 767: Only the payment path is covered for pending line items. Shall we add the disbursement path - it doesn't go through `mapLineItemsToDisplayItems` and instead passes the raw `type` field straight into `additionalLineItems`

```suggestion
  });

  it("passes pending line item type through additionalLineItems on disbursement sessions", async () => {
    await buildSession(applePay, {
      transaction: {
        ...disbursementTransaction,
        lineItems: [
          { label: "Item", amount: 1000 },
          { label: "Estimated tax", amount: 80, type: "pending" },
        ],
      },
    });

    const modifierData = (
      paymentRequestCalls[0].modifiers?.[0] as unknown as {
        data: { additionalLineItems: Array<Record<string, unknown>> };
      }
    ).data;

    expect(modifierData.additionalLineItems).toEqual([
      { label: "Total Amount", amount: 1000 },
      { label: "Item", amount: { value: "10.00", currency: "USD" } },
      {
        label: "Estimated tax",
        amount: { value: "0.80", currency: "USD" },
        type: "pending",
      },
      {
        label: "Apple Pay Demo",
        amount: 1000,
        disbursementLineItemType: "disbursement",
      },
    ]);
  });
```

ana-maksimovskikh (approved):
Thank you @Jason-Papa - looks great & good to go 🎉 

All comments are optional, feel free to reject if you feel it's not needed 😄

packages/browser/lib/ui/ApplePay/index.ts
Jason-Papa: Added

packages/browser/lib/ui/ApplePay/utilities.ts
Jason-Papa: I think its not exactly the same, leaving as is for now

packages/browser/lib/ui/ApplePay/utilities.ts
```diff
@@ -25,6 +25,14 @@ import {
 import ApplePayButton from ".";
 import { RecurringPaymentIntervalUnit } from "types/uiComponents";
 
+/**
+ * Probe ceiling for `ApplePaySession.supportsVersion()`.
+ *
+ * Keep in sync with the major version of `@types/applepayjs` — a unit test fails
+ * if that package moves ahead of this constant.
+ */
+export const APPLE_PAY_MAX_VERSION = 14;
```
alexander-rw, line 34: praise: this is a big 🧠 idea , like the fact there's a test that causes this to fail

packages/browser/lib/ui/ApplePay/utilities.ts
```diff
@@ -86,12 +96,75 @@ function applyContactFields(
   }
 }
 
+function assertBase64ApplicationData(applicationData: string) {
+  if (typeof applicationData !== "string") {
+    throw new Error("applicationData must be a Base64-encoded string");
+  }
+
+  try {
+    atob(applicationData);
```
alexander-rw, line 105: should this be `return atob(…)` ?

alexander-rw (commented):
some minor comments, not blocking

packages/browser/lib/ui/ApplePay/utilities.ts
alexander-rw: or are we just checking if we can actually do the atob and that's enough

PR #968, opened 2026-08-03
Fix duplicate 3DS iframe post guards

jmorgadosoares:
# Why
[EXP-1251](https://linear.app/evervault/issue/EXP-1251/fix-duplicate-3ds-iframe-post-guards)

In `BrowserFingerprint.tsx` and `ChallengeFrame.tsx`, the guard that prevents a duplicate post to the iframe only keys on one field (`data` or `creq`), while the actual post also depends on `url`. If a retry comes in with the same `data`/`creq` but a new `url`, the guard sees no change and skips the post, so the frame never posts to the new url.

The ticket also raised whether `action.data` could be `null` in a valid flow. It can't: the backend only emits `type: "browser-fingerprint"` as the next action when there's fingerprint data to post, so there's no valid 3DS flow where this action exists with empty/null data. Adding a null check here would be defending against a scenario that can't occur given the contract between the session response and this component, and if the API response were ever malformed that's better caught when the response is parsed, not deep inside this component. So no change was made for that part.

# How
Replaced the single-field refs (`postedData`, `postedCreq`) with a ref that holds the full next action object (`BrowserFingerprintNextAction` / `ChallengeNextAction`), and compare both the relevant data field and `url` before skipping the post. A retry with either field changed now posts again.


ana-maksimovskikh (approved):
Thanks @jmorgadosoares 🙏 
LGTM 🔥

PR #970, opened 2026-08-04
Fix unguarded postMessage listeners 

ana-maksimovskikh:
Bug fix: https://linear.app/evervault/issue/CARD-1087/unguarded-postmessage-listeners-in-evervaultframe-usemessaging-can


Jason-Papa (approved):
lgtm

PR #977, opened 2026-08-12
Convert Google Pay amounts using the currency's exponent

maxharrison:
`buildPaymentRequest` always divided minor-unit amounts by `100`, so Google Pay showed zero-decimal currencies at one hundredth of their value and three-decimal currencies at ten times their value.

It now formats totals and line items with each currency's exponent and **rejects Google Pay amounts that require a third fractional digit**, because **Google Pay accepts at most two**. This corrects the documented minor-unit contract, but merchants that compensated for the old formatting must remove that workaround, so this change is held for the next major release.

### DIFF
--- a/.changeset/google-pay-currency-exponent.md
+++ b/.changeset/google-pay-currency-exponent.md
@@ -1,9 +1,20 @@
 ---
-"@evervault/ui-components": patch
+"@evervault/ui-components": minor
 ---
 
 Fix Google Pay amounts for currencies that do not have two minor units. Amounts
 were always divided by 100 and rendered with two decimal places, so a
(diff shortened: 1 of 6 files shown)


ana-maksimovskikh (approved):
Good catch! Thanks, Max 🙏

maxharrison:
Updated the changeset to `major` and added coverage for a zero JPY total when `lineItems` is omitted

maxharrison:
This PR corrects how minor units are handled of non-2-decimal currencies, but merchants that compensated for the old formatting must remove that workaround, so this change is held for the next major release.

ana-maksimovskikh:
Let's maybe let it hang for a bit, it's not an urgent change
And would be a good one to test new release flow once it's ready 👀

PR #978, opened 2026-08-12
Convert Apple Pay amounts using the currency's exponent

maxharrison:
Apple Pay request builders always divided minor-unit amounts by `100`, so zero-decimal currencies rendered at one hundredth of their value and three-decimal currencies at ten times their value. They now use `formatTransactionAmount` for payment and recurring totals, line items, shipping options, payment-method updates, and disbursement line items, using the currency exponent supplied by `Intl.NumberFormat`. This corrects the documented minor-unit contract, but merchants that compensated for the old formatting must remove that workaround, so this change is held for the next major `@evervault/browser` release.

Note: This PR is stacked on #977.

### DIFF
--- a/packages/browser/lib/ui/ApplePay/utilities.ts
+++ b/packages/browser/lib/ui/ApplePay/utilities.ts
@@ -1,4 +1,5 @@
 import { getAppSDKConfig } from "shared/getAppSDKConfig";
+import { formatTransactionAmount } from "shared/currency";
 import {
   ApplePayMerchantCapability,
   ApplePayTransactionType,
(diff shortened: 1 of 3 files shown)


packages/browser/lib/ui/ApplePay/utilities.ts
ana-maksimovskikh, line 883: Should we also apply the format here?

ana-maksimovskikh (changes requested):

ana-maksimovskikh:
We might have an issue with Google rejecting formatted string on their side - worth doublechecking

https://developers.google.com/pay/api/web/reference/request-objects#TransactionInfo

maxharrison:
Closing this draft. We will keep the Apple Pay correction separate from the Google Pay major-release work.

PR #985, opened 2026-08-20
Add Google Pay golden request tests

maxharrison:
The web Google Pay request had no whole-request test, so defaults and configured fields could change without a focused failure. `googlePayGoldenRequest.test.ts` now compares billing-disabled, billing-enabled, and custom requests with JSON fixtures.

The fixture type derives from `buildPaymentRequest`, so the projection stays tied to the production return type. A separate test covers web-only `merchantId`, `merchantOrigin`, and `callbackIntents` fields.

The fixtures match the Android SDK copies byte-for-byte, but no CI check enforces that yet.

Stacked on #987.

### DIFF
--- a/packages/ui-components/test/googlePayGoldenRequest.test.ts
+++ b/packages/ui-components/test/googlePayGoldenRequest.test.ts
@@ -1,0 +1,129 @@
+import { readFileSync } from "node:fs";
+import { describe, expect, it } from "vitest";
+import { MerchantDetail } from "types";
+import { buildPaymentRequest } from "../src/GooglePay/utilities";
+import { GooglePayConfig } from "../src/GooglePay/types";
+import { apiConfig } from "../src/utilities/config";
+
+/**
+ * Fixture tests for the whole Google Pay payment request.
+ *
+ * The fixtures are shared with the Android SDK's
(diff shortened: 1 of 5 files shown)


packages/ui-components/test/googlePayGoldenRequest.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    );
+  });
+
+  it("matches the custom fixture when every option is set", () => {
+    expect(
+      comparableRequest({
+        ...baseConfig,
+        transaction: { ...transaction, priceLabel: "Subscription" },
+        emailRequired: true,
+        allowedAuthMethods: ["PAN_ONLY"],
```
ana-maksimovskikh, line 113: nit: no test exercises `allowedAuthMethods/allowedCardNetworks` independently of `billingAddress` in the "custom" case

ana-maksimovskikh (approved):
Thanks @maxharrison - the change looks good 🙌

packages/ui-components/test/googlePayGoldenRequest.test.ts
maxharrison: Will follow up in a later PR

PR #986, opened 2026-08-20
Omit Google Pay billingAddressParameters when billing is not required

maxharrison:
Stacked on #985. Review that first.

`buildPaymentRequest` always sent `billingAddressParameters`, even when `billingAddressRequired` was false:

```json
"billingAddressRequired": false,
"billingAddressParameters": { "format": "FULL", "phoneNumberRequired": false }
```

Google ignores the parameters when `billingAddressRequired` is false, so the payment sheet does not change. The request still claimed a `FULL` address format that we never asked for, which is misleading in logs and in support tickets.

Omitting the field makes the request state only what it means, and makes `billing-disabled.json` identical to the Android fixture.

Not a breaking change for merchants. The only visible effect is one fewer field on the wire, shown here as the fixture diff.

### DIFF
--- a/packages/ui-components/src/GooglePay/utilities.ts
+++ b/packages/ui-components/src/GooglePay/utilities.ts
@@ -30,10 +30,17 @@ export function buildPaymentRequest(
               "VISA",
             ],
           billingAddressRequired: isBillingRequired(config),
-          billingAddressParameters: {
-            format: billingAddressFormat(config),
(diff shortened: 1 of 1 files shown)


maxharrison:
Reopened as #987 after the stack was reordered. GitHub auto-closed this one as merged into its former stack base; nothing landed on `master`.

PR #987, opened 2026-08-20
Omit Google Pay billingAddressParameters when billing is not required

maxharrison:
`buildPaymentRequest` sent `billingAddressParameters` even when `billingAddressRequired` was `false`. Google ignores those parameters, but the request still claimed a `FULL` billing address format that the merchant did not request.

The request now omits `billingAddressParameters` unless billing collection is enabled. This does not change the Google Pay sheet, and it matches the Android request for the disabled case.

Stack: #985 adds golden request tests on top of this PR.

### DIFF
--- a/.changeset/google-pay-button-radius-default.md
+++ b/.changeset/google-pay-button-radius-default.md
@@ -1,0 +1,5 @@
+---
+"@evervault/ui-components": minor
+---
+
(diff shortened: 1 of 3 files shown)


maxharrison:
Should we release a changeset with this?

ana-maksimovskikh (approved):
LGTM - thank you 🙏

PR #990, opened 2026-08-25
ci: accept main as well as master in workflow triggers

aevv:
Preflight for the org-wide default branch rename to `main`.

Workflow triggers now match both `master` and `main`, so nothing stops
firing during the rename. The `master` entries are removed in the follow-up
cleanup PR after the default branch has been flipped.

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -5,6 +5,7 @@ on:
     branches-ignore:
       - master
       - main
+      - main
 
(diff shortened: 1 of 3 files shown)


aevv:
Closing as a no-op: this repo's workflow triggers already accept both `master` and `main`, so no preflight is needed before the default branch rename. The diff here was a duplicate entry produced by a bug in the migration script, since fixed.

PR #1002, opened 2026-09-02
Prevent production releases blocking master pushes

maxharrison:
The workflow-wide `concurrency` group held later pushes while a production release waited for approval. Build artifacts also expired after seven days, which prevented delayed releases from deploying.

The workflow now sets job-level concurrency, retains build artifacts for 30 days, and posts a Slack notification when a production release awaits approval.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,6 +68,7 @@ jobs:
         uses: actions/upload-artifact@v7
         with:
           overwrite: false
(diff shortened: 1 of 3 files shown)


ana-maksimovskikh:
@maxharrison can we just remove the manual release confirmation bit  🙏 
Since a release PR gets manually approved anyway before push, that seems obsolete 👀

PR #1007, opened 2026-09-07
Add Google Pay offers to the web sheet

maxharrison:
Google Pay on web declares no `offerInfo` and no `OFFER` callback intent, so a buyer cannot apply a promotion in the sheet even though the Google Pay Web API supports one. This declares the merchant's `offers` on the request, adds `OFFER` to the derived `callbackIntents`, and routes the `OFFER` trigger to a merchant `onOfferChange` callback that returns an updated total, line items or offer list; the applied codes are surfaced on the `process()` payload as `redemptionCodes`. Promotion entry now works in the web sheet, and an offer the merchant rejects defaults to the `OFFER_INVALID` reason and `OFFER` intent Google expects rather than a shipping reason.

Stacked on #1006 — it reuses the correlated data-change handshake added there, which is what keeps this change small. Review and merge that first; this diff is only the offer-specific parts.

Closes CARD-1058.

### DIFF
--- a/.changeset/google-pay-web-offers.md
+++ b/.changeset/google-pay-web-offers.md
@@ -1,0 +1,6 @@
+---
+"@evervault/browser": minor
+"@evervault/js": minor
+---
+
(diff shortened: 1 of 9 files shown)


maxharrison:
Gonna cancel this ticket. The feature is not available on Google Pay Android so will revisit later.

PR #1008, opened 2026-09-07
Share the Apple Pay SDK load promise across ui.applePay() instances

alexander-rw:
Closes CARD-1239. A second `ui.applePay()` instance on the same page probed Apple's SDK before it had executed and then memoized the resulting `"unsupported"` forever, so the express-checkout button never mounted — the load promise is now shared across instances rather than per-instance, and only `"available"` is cached.

Reported by a customer via solutions-engineering; it reached CDN users on Sep 2 when `js.evervault.com/v2` finally served 2.63.0.

### DIFF
--- a/.changeset/apple-pay-shared-sdk-load.md
+++ b/.changeset/apple-pay-shared-sdk-load.md
@@ -1,0 +1,5 @@
+---
+"@evervault/browser": patch
+---
+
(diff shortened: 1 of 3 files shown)


alexander-rw:
Superseded by #1012, which carries the same fix for CARD-1239. Closing this duplicate; #1010 stacks on #1012.

PR #1010, opened 2026-09-08
fix(browser): surface a failed Apple Pay credentials exchange on the error event

alexander-rw:
A failed Apple Pay credentials exchange currently dies as an uncaught `TypeError` with no error event and no sheet dismissal, this fixes it.


alexander-rw:
Not actually merged to `master`. GitHub auto-marked this merged during a restack, and apparently it has a UI bug. actual pr is 1012

PR #1013, opened 2026-09-08
fix(browser): surface a failed Apple Pay credentials exchange on the error event

alexander-rw:
Closes CARD-1247. A failed Apple Pay credentials exchange currently dies as an uncaught `TypeError` with no error event and no sheet dismissal, which is why the Upgrow outage ran silent for five days — this puts the failure on the SDK's error channel where a merchant can actually see it.

#1012 stacks on this. Replaces #1010, which GitHub auto-marked merged when a restack briefly made its head an ancestor of its base.

### DIFF
--- a/.changeset/apple-pay-credentials-error-handling.md
+++ b/.changeset/apple-pay-credentials-error-handling.md
@@ -1,0 +1,5 @@
+---
+"@evervault/browser": patch
+---
+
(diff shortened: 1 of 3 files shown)


alexander-rw:
Folded into #1012 — the res.ok commit `da533aa4` is already in that branch's history, and #1012 now targets `master` directly as a single PR. Closing this rather than keeping the stack.

PR #1017, opened 2026-09-08
chore(dep-mgmt): override vulnerable versions of brace-expansion

lfarrel6:
# Why

Brace expansion has vulnerabilities in both v1 and v2. Scope is limited as the package is a transitive dep and no uncontrolled inputs are being passed directly to the lib.

# How

Override vulnerable versions of brace-expansion


alexander-rw (approved):

evervault-dependencies:
This vulnerability has been resolved through the standard development process and is no longer reported by Vanta. Closing this PR automatically.

PR #1027, opened 2026-09-15
Publish every SDK release to a version-pinned CDN path

jmorgadosoares:
# Why

We cannot currently pin an Evervault JS SDK to a version. Every customer loads whatever is live at `js.evervault.com/v2`, `inputs.evervault.com/v2`, `ui-components.evervault.com` and `3ds.evervault.com`, so a bad release reaches everyone at once and customers have no way to hold a known-good build.

This PR publishes every release to a version-pinned path alongside the live one. The companion PR in `devops` adds the CloudFront behaviors that serve those paths.

https://github.com/evervault/devops/pull/4134

# How

## Pinned uploads

Each S3 deploy step now also uploads the same artifact to a version prefix derived from the release tag, before it touches the live path, so a failure there leaves live untouched:

| Package | Live path | Pinned path |
| --- | --- | --- |
| `@evervault/browser` | `/v2/` | `/v2@<version>/` |
| `@evervault/inputs` | `/v2/` | `/v2@<version>/` |
| `@evervault/ui-components` | `/` | `/@<version>/` |
| `@evervault/3ds` | `/` | `/@<version>/` |

Pinned objects upload with a one year `immutable` `Cache-Control`.

## Write-once

A pinned version that is already published must never be overwritten, otherwise pinning means nothing. A shared `assert_unpublished` helper checks the prefix with `aws s3api list-objects-v2`, which exits zero with no keys when the prefix is free and non-zero only on a real error. A check that cannot run blocks the publish rather than reading as "not published yet". If a version is already there the job fails and asks for a new patch version.

## Relative asset paths

`ui-components`, `inputs` and `3ds` ship an HTML entry point that references hashed assets. Those builds now use `base: "./"` so one build works under both the live prefix and the pinned one, with no separate build per path. Two things follow:

- `inputs` passes `publicPath: "./"` to `rollup-plugin-sri` so it keeps finding each file in the bundle, and the `ui-components` integrity plugin strips a leading `./` as well as a leading `/`. Verified that SRI hashes are still emitted on every script and stylesheet in all three packages.
- The `ui-components` preview needs its own copy of the assets under `/preview/assets` instead of sharing the live `/assets`. Pinned uploads publish assets before `index.html` so the entry point is never reachable ahead of the files it references.

## Staging coverage

Pushes to `main` pass a throwaway `0.0.0-staging-<run_id>-<run_attempt>` tag so the pinned upload path runs on every merge, not only on releases. The devops PR expires those objects after 30 days.

## Testing

All three affected packages were rebuilt and their output inspected: relative asset references, SRI hashes intact, no absolute `/assets` references and no dynamic imports that would break under a relative base. The write-once guard was run against stubbed free, occupied and `AccessDenied` responses.


ana-maksimovskikh:
@jmorgadosoares Not sure how valid this concern is, but we might be required to serve the lates version for PCI compliance - so we might not want to make all the previous versions publickly available 

For ApplePay Web SDK docs have the following:

>  "The SDK must be loaded directly from our CDN and cannot be bundled with your application or self hosted."
> 
> "The loadEvervault function will always load the latest version of the Evervault SDK regardless of which version of @evervault/js you are using."

The docs explicitly frame CDN-loading as required for PCI compliance, not just a convenience.

A bit more about it in here: https://evervault.slack.com/archives/C0116FH4B53/p1788947682969729

jmorgadosoares:
Super valid concern! Thanks for reviewing

I didn't realise the scope of the requirements for something like this so I'll close this PR until I have a better understanding

PR #1029, opened 2026-09-17
feat: add fontFaces theme option for self-hosted brand fonts

AnyaDroz:
## what
- adds a font faces option: customers pass their own font as base64 in the config, instead of the google fonts allowlist we have now
- mitigates visual spoofing (a font drawing a digit wrong, eg 4 as 9) by rejecting blank or reused digit glyphs
- there is acsp change needed for these embedded fonts: evervault/devops#4143

## how
- validates the font source is a strict base64 embed of an allowed type (woff2, woff, ttf, otf), rejects remote urls, other file types, css-injection attempts
- checks digits 0-9 via a font-parsing library: each needs a real drawing, not the empty missing-character placeholder, and no two digits sharing one
- library loads lazily, only when a theme sets custom font faces, keeps it off the main iframe bundle (its needed for WOFF2 🐶)
- skipped full visual verification (render + compare each digit's shape to a reference, see https://tritium.legal/blog/noroboto), too heavy, risks rejecting legit stylized fonts for a digits-only field (but cool article!)
- checked compliance: nothing in pci covers font content, confirmed against the standard text
- pci 6.4.3 covers scripts, not fonts: "a method is implemented to confirm that each script is authorised... assure the integrity of each script" - not applicable, but (do we do this now for our actual scripts?)
- pci 11.6.1 requires tamper-detection on "security-impacting HTTP headers... of payment pages as received by the consumer browser" - that's the csp change (devops#4143), not this pr (do we do this now?)

context:
- https://evervault.slack.com/archives/C09KNHABETT/p1789174210722639 (customer ask)
- https://evervault.slack.com/archives/C0BEHR7T57V/p1789390859503929 (internal thread)

## usage
```ts
theme: {
  fontFaces: [
    {
      fontFamily: "Natural Sans",
      src: "data:font/woff2;base64,d09GMgABAAAA...",
      fontWeight: 400,
    },
  ],
  styles: {
    ".Input": { fontFamily: "Natural Sans" },
  },
}
```

PR #1032, opened 2026-09-18
fix(ui-components): hash emitted files for SRI so code split builds match

AnyaDroz:
## What 

ui-components iframe carries a feature called [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Subresource_Integrity) - this checks that resources we fetch haven't been tampered with. 

Lazy loading fontkit broke that check - I have more work coming that splits ui-components into separately loaded modules. [the related change](https://github.com/evervault/evervault-js/pull/1029)

## How this PR fixes this 

The plugin was taking the 'fingerprint' from the build tools in memory copy of the file, while the HTML was being generated. The build was changing that file with the lazy load before writing it to disk. Vite emits a placeholder when compiling dynamic imports, so we just need to wait to run the SRI until the build has written the files, where vite has subbed the placeholder. 

[My follow up PR stacked on top of here](https://github.com/evervault/evervault-js/pull/1032) will check the integrity of module-split imports. No PCI is affected as discussed with others; this is not classified as a payments portal.


AnyaDroz:
@ev-vaultkeeper review

packages/ui-components/vite/integrity.ts
```diff
@@ -1,26 +1,39 @@
...
         const cleaned = src.startsWith("/") ? src.slice(1) : src;
-        const resource = ctx.bundle?.[cleaned] as { code?: string };
 
-        if (!resource?.code) return;
-        const hash = generateIntegrity(resource.code);
-        node.setAttribute("integrity", hash);
+        let code;
+        try {
+          code = readFileSync(resolve(outDir, cleaned));
+        } catch {
```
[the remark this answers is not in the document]
AnyaDroz: matches existing

packages/ui-components/vite/integrity.ts
```diff
@@ -1,26 +1,39 @@
...
     apply: "build" as const,
 
-    transformIndexHtml(html: string, ctx: IndexHtmlTransformContext) {
-      const parsed = new jsdom.JSDOM(html);
+    configResolved(config: ResolvedConfig) {
+      outDir = resolve(config.root, config.build.outDir);
+    },
+
+    closeBundle() {
+      const indexPath = resolve(outDir, "index.html");
```
ana-maksimovskikh, line 21: Might bite us later if the build ever grows to more than one page

We could instead collect emitted .html assets from the bundle (e.g. in generateBundle, alongside the chunk tracking added in [evervault/evervault-js#1033](https://github.com/evervault/evervault-js/pull/1033)) and loop over all of them here, so this stays correct without relying on an assumption about the entry count

Not sure, how much this needed though

packages/ui-components/vite/integrity.ts
ana-maksimovskikh, line 32: For a security control I'd lean toward failing loud here: throw new Error(\failed to read ${cleaned} for SRI: ...`) instead of return`, so a broken path gets caught in CI rather than silently shipping an unprotected script. 

Or was fail-open intentional here, e.g. to tolerate external/CDN-ish local paths that don't need hashing?

ana-maksimovskikh (approved):
Approved with two comments 👀 

Probably good to go as is, but thought better to mentioned

packages/ui-components/vite/integrity.ts
AnyaDroz: yes I saw the bot mention this, so, this branch is unreachable as of now so single page, two files, both emitted by every build. It's also pre-existing behaviour rather than something this PR introduces. Leaving it as-is here to unblock the release, and hardening it in #1033, which restructures this same function and is where the file count grows 👍

packages/ui-components/vite/integrity.ts
AnyaDroz: A second html isn't planned so I'll leave as is for now, thanks!!

PR #1033, opened 2026-09-18
feat(ui-components): add import map integrity for dynamically imported chunks

AnyaDroz:
Using [importMap](https://shopify.engineering/shipping-support-for-module-script-integrity-in-chrome-safari) - Shopify made this happen where the fingerprints are declared up front as a URL to hash against instead of being bound to an element.


AnyaDroz:
@ev-vaultkeeper review

packages/ui-components/vite/integrity.ts
```diff
@@ -50,6 +63,30 @@ export function integrity() {
...
+        if (referenced.has(fileName)) continue;
+        dynamic[`/${fileName}`] = generateIntegrity(
+          readFileSync(resolve(outDir, fileName))
+        );
+      }
+
+      if (Object.keys(dynamic).length > 0) {
+        const map = parsed.window.document.createElement("script");
+        map.setAttribute("type", "importmap");
+        map.textContent = JSON.stringify({ integrity: dynamic });
```
ana-maksimovskikh, line 86: Browser support for the integrity key inside import maps is still limited
In browsers that don't support it, this field is just ignored, so dynamically-imported chunks get silently zero enforcement there. 
Worth being explicit (e.g. a code comment) that this doesn't uniformly close the gap yet

packages/ui-components/vite/integrity.ts
```diff
@@ -50,6 +63,30 @@ export function integrity() {
...
+        parsed.window.document.head.prepend(map);
```
ana-maksimovskikh, line 87: There can only be one importmap per document, and it must be inserted before any module script has executed. 

if closeBundle ever fires more than once for the same HTML file (e.g. multiple rollup outputs), this would prepend a second importmap and the browser would reject/ignore it. 

Given [evervault/evervault-js#1032](https://github.com/evervault/evervault-js/pull/1032)'s chunk-tracking already had to be fixed for "accumulate across rollup outputs," is multi-output something this build could hit, or is single-output guaranteed long-term?

ana-maksimovskikh (approved):
Thanks Anya, LGTM

Added comments with a few smaller concerns, but don't think they are properly blocking

packages/ui-components/vite/integrity.ts
AnyaDroz: [I did read](https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Subresource_Integrity) about this before making the PR - but I'd rather avoid a comment about something that will drift (when they add support) - this PR is adding 88% coverage as opposed to creating new risk, so I'll just make sure this is named here in the thread for future investigators 👍

packages/ui-components/vite/integrity.ts
AnyaDroz: I added a guard! thanks 🙏

PR #1037, opened 2026-09-18
<ev-card> renders the default card via a registered custom element

jmorgadosoares:
# Why
Walking skeleton for Phase 2 (DOM composability): `<ev-card>` renders the default card through a registered custom element, without going through `evervault.ui.card()`.

Closes EXP-1491 (with EXP-1497, EXP-1499, EXP-1500, EXP-1501, EXP-1502). Also carries the pieces of EXP-1492 and EXP-1521 the element cannot mount without: the spec node types, `serialise()`, patch application and the renderer reading one node tree.

RFC: [Phase 2 - DOM composability and customisability](https://app.notion.com/p/evervault/Phase-2-DOM-composability-and-customisability-3d6d75cfe0ca8102876cd8706a95503e)

# How
`ui.card()` and `<ev-card>` become sibling front-ends over one shared core, `CardFrame`, which owns the iframe lifecycle, the message subscriptions and their release, `values` and `validate()`. `ui.card()` keeps its public surface and only translates `CardOptions`.

```html
<script src="https://js.evervault.com/v2"></script>
<ev-card team-id="team_..." app-id="app_..."></ev-card>
```

```js
// or mount every <ev-card> on the page with an instance
const evervault = new Evervault(TEAM_ID, APP_ID);
evervault.ui.mount();

document.querySelector("ev-card").addEventListener("change", (event) => {
  event.detail; // same payload as card.on("change")
});
```

The renderer normalises `config.fields` at its boundary: a legacy `CardField[]` from `ui.card()` becomes the fixed-order tree, a `CardSpecNode[]` from `<ev-card>` is rendered as written. Everything downstream reads nodes. The legacy translation can never be removed, since the renderer is one rolling deployment loaded by every SDK version ever shipped; a 200-sample recording of the pre-change renderer (`test/fixtures/cardLegacyFields.jsonl`) pins it.

```ts
// packages/types/uiComponents.ts
export interface CardFrameConfig {
  fields?: CardField[] | CardSpecNode[];
  // ...
}
```

The element mounts the frame into a closed shadow root with no slot, so light-DOM children stay inert. Removing the element destroys the frame and releases every window listener; a test counts them.

[screenshot of `<ev-card team-id app-id>` rendering the default number, expiry and cvc fields]

Not in this PR: reading declared children (EXP-1492 PR), field attributes, auto-advance and card-level attributes (EXP-1493 PR).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jmorgadosoares:
Superseded: this PR is split into #1042 (card frame core), #1043 (renderer node tree), #1044 (its tests) and a new PR for the element itself. GitHub does not allow changing the base of a PR in a stack, hence the new PR.

jmorgadosoares:
@ev-vaultkeeper review

PR #1038, opened 2026-09-18
Serialize declared children into the frame contract

jmorgadosoares:
# Why
`<ev-card>` renders what its children declare, and follows changes to them without remounting or losing typed input.

Closes EXP-1492 (with EXP-1503, EXP-1504, EXP-1505, EXP-1506). Stacked on the EXP-1491 PR.

# How
The element serialises its light-DOM children into spec nodes and hands the tree to the core. `CardFrame` tracks the tree the frame holds, diffs against it and sends only the difference as `EV_SPEC_PATCH`; the renderer applies the patch by stable id, so untouched fields keep their values.

```html
<ev-card>
  <ev-card-number placeholder="Card number"></ev-card-number>
  <ev-row>
    <ev-card-expiry></ev-card-expiry>
    <ev-card-cvc></ev-card-cvc>
  </ev-row>
</ev-card>
```

```ts
// serialised
[
  { type: "number", id: "ev-1", props: { placeholder: "Card number" } },
  { type: "row", id: "ev-2", props: {}, children: [
    { type: "expiry", id: "ev-3", props: {} },
    { type: "cvc", id: "ev-4", props: {} },
  ]},
]
```

A `MutationObserver` coalesces DOM changes into one read per microtask. Unknown children are ignored with a warning. Declaring nothing keeps the default card; removing every child returns to it.

[screenshot of the declared card above, number on its own line and expiry next to cvc]

[screenshot of the console after adding a child at runtime, showing one `EV_SPEC_PATCH` with a single insert]

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jmorgadosoares:
@ev-vaultkeeper review

